PR #2, opened 2016-03-22
added first proposal of EThcD term

mwalzer:
after a quick exchange with Thermo development it turned out such a term would be needed for exporting mzML in given acquisition scenarios.
fixes issue https://github.com/HUPO-PSI/psi-ms-CV/issues/3


germa:
Term already added as
id: MS:1002631
name: Electron-Transfer/Higher-Energy Collision Dissociation (EThcD)

mwalzer:
Yeah, this was part of my proposal, how suggestions for new terms could be made by (CV versed) users in a more github like manner. See also https://github.com/HUPO-PSI/psi-ms-CV/releases/tag/v3.83.0-showcase.github.release

PR #35, opened 2020-07-15
add 5 ion interpretation format terms

edeutsch:
As proposed in Peak interpretation format - Peptides:
https://docs.google.com/document/d/1yEUNG4Ump6vnbMDs4iV4s3XISflmOkRAyqUuutcCG2w

### DIFF
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -1,6 +1,6 @@
 format-version: 1.2
-data-version: 4.1.38
-date: 30:06:2020 14:30
+data-version: 4.1.39
+date: 14:07:2020 14:30
 saved-by: Eric Deutsch
 auto-generated-by: OBO-Edit 2.3.1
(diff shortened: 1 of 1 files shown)


psi-ms.obo
```diff
@@ -20123,6 +20123,36 @@ def: "Term for a comment field withing the NIST msp file format" [PSI:MS]
...
+
+[Term]
+id: MS:1003104
+name: peptide ion interpretation format
+def: "Interpretation format designed primarily for peptides, with allowances for generic chemical formulas and other miscellaneous named ions." [PSI:MS]
+is_a: MS:1003103 ! ion interpretation format
+
+[Term]
+id: MS:1003105
+name: cross-linked peptide ion interpretation format
```
mwalzer, line 20140: How does that compare with XL-MOD? https://github.com/HUPO-PSI/mzIdentML/blob/master/cv/XLMOD.obo

PR #36, opened 2020-07-17
PatternLab information

ypriverol:
Some scores and terms for PatternLab software (http://www.patternlabforproteomics.org/)


ypriverol:
@edeutsch done!!!!

PR #37, opened 2020-07-30
Add Orbitrap ID-X

chambm:
Does this look good @jshofstahl @edeutsch ?


chambm:
There we go. CI approved! :100:  Now we can do simple additions directly from GitHub and not worry about making the document invalid (although that duplicate ID thing still needs checking: I might ask fastobo-validator to add a warning for that which is optionally fatal).

PR #39, opened 2020-07-30
Create validate-obo.yml

chambm:


chambm:
Well I guess it validates? Let's try breaking it.

chambm:
Interesting. I didn't know this: `Although it is strongly recommended that an OBOF document does not contain two frames with the same identifier, this is syntactically valid. If two frames have the same identifier, then they are combined. Only frames of the same type can be combined - if a document uses the same ID for two frames of different types, the document is structurally invalid.`

edeutsch:
> Interesting. I didn't know this: `Although it is strongly recommended that an OBOF document does not contain two frames with the same identifier, this is syntactically valid. If two frames have the same identifier, then they are combined. Only frames of the same type can be combined - if a document uses the same ID for two frames of different types, the document is structurally invalid.`

thanks for mentioning this. indeed interesting/alarming.

PR #40, opened 2020-08-24
OpenMS added score

ypriverol:
@timosachsenberg @julianus new term  `OpenMS: ConsesusID PEP` created

### DIFF
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -20211,7 +20211,7 @@ is_a: MS:1001143 ! PSM-level search engine specific statistic
 [Term]
 id: MS:1003116
 name: OpenMS:Target-decoy peptide q-value
-def: "The OpenMS Target-decoy q-values at Peptide sequence level" [PSI:MS]
+def: "The OpenMS Target-decoy q-values at peptide sequence level" [PSI:MS]
 xref: value-type:xsd\:double "The allowed value-type for this CV term."
 is_a: MS:1002358 ! search engine specific peptide sequence-level identification statistic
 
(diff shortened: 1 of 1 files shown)


ypriverol:
Thanks, @edeutsch for your comment. I have discussed it with the OpenMS team and we changed the meaning. However, we have now another term for that actually is at three levels because is computed in the same way `OpenMS:Target-decoy q-value`

edeutsch:
Thank you for changing the ConsensusID PEP term. It was only a PSM-level score after all, I guess?

However, I am now concerned about the `OpenMS:Target-decoy q-value` term for the same reason. It may be "computed in the same way", but surely using different numbers and the results (the term) has a different semantic meaning if computed at the PSM level vs. computed at the peptide-sequence level vs. computed at the protein level. I would discourage using the same term for three different semantic concepts. I think it would be better to have 3 terms for 3 different semantic concepts:
`OpenMS:Target-decoy PSM-level q-value`
`OpenMS:Target-decoy peptide-sequence-level q-value`
`OpenMS:Target-decoy protein-level q-value`
each with different parentage.

But If you all feel differently, I can be outvoted.

chambm:
The scores at different levels are never comparable so it doesn't make sense for them to be the same term.

ypriverol:
LEts wait for the OpenMS guys for their comments and we can merged it.

psi-ms.obo
```diff
@@ -20183,9 +20184,58 @@ is_a: MS:1001139 ! quantitation software name
...
+id: MS:1003115
+name: OpenMS:Target-decoy PSM q-value
+def: "The OpenMS Target-decoy q-values at PSM level" [PSI:MS]
+xref: value-type:xsd\:double "The allowed value-type for this CV term."
+is_a: MS:1001143 ! PSM-level search engine specific statistic
+
+[Term]
+id: MS:1003116
+name: OpenMS:Target-decoy peptide q-value
+def: "The OpenMS Target-decoy q-values at Peptide sequence level" [PSI:MS]
```
timosachsenberg, line 20214: should peptide be lower-case? see protein below

timosachsenberg (approved):
look good

PR #41, opened 2020-09-01
Terms for OpenMS created. 

ypriverol:


timosachsenberg:
loos good. Thanks for adding.

PR #43, opened 2020-09-07
add value type for FAIMS compensation voltage

timosachsenberg:


edeutsch (approved):
Looks like a good fix. I wonder why this wasn't already there. Do you suppose it should have units as well? Or is the units always just volts and thus unnecessary?

timosachsenberg:
I think it is volts UO:0000218 and we could add it as well

edeutsch (changes requested):
Looks great. Would you also bump the version number and date at the top of the file?

edeutsch (approved):
Great, cleared to merge IMO
@ypriverol or anyone else have concerns?

ypriverol (approved):

PR #45, opened 2020-09-25
add Bruker instrument rapiflex

ypriverol:


ypriverol:
@edeutsch csn you have a look?

edeutsch (approved):
Looks good to me. Only potential quibble is that Bruker's marketing materials capitalizes the name as rapifleX with a capital X. Seems like it would be nice to preserve that properly. Although admittedly not done correctly in other Bruker instrument names. But why not do it correctly for new instruments at least?

ypriverol:
I will change it now.

PR #46, opened 2020-10-19
timsTOF fleX instrument added 

ypriverol:


edeutsch (changes requested):
I think a better parent would be:
is_a: MS:1001534 ! Bruker Daltonics flex series

What do you think?

edeutsch (changes requested):
also, each change should increment the data-version number

chambm:
Probably need a new parent for timsTOF instruments. Because I think the only thing it shares in common with the flex series is the MALDI source (which is probably very different, MALDI-2 apparently).

edeutsch:
Yes, that sounds sensible. So add a new term:

Bruker Daltonics timsTOF series

and this proposed term and also previous term:
id: MS:1003005
name: timsTOF Pro

under this new parent, too?

ypriverol:
@edeutsch @chambm I did the requested changes.

chambm (approved):

edeutsch (approved):
Looks great, thanks! Ready to merge!

PR #50, opened 2021-01-18
Demonstrate encoding value types with relationships

cthoyt:
Many terms seem to be annotating their types using formations like this:
```
[Term]
id: MS:1000005
name: sample volume
def: "Total volume of solution used." [PSI:MS]
xref: value-type:xsd\:float "The allowed value-type for this CV term."
is_a: MS:1000548 ! sample attribute
relationship: has_units UO:0000098 ! milliliter
```

This isn't typically how xrefs are used. I think it would be better to create a new `[Typedef]` for `value-type` that would allow a relationship like

`relationship: has_value_type xsd:float`

I accidentally opened an issue on the PSI-MI controlled vocabulary (https://github.com/HUPO-PSI/psi-mi-CV/issues/434) where I had a short discussion with @pporrasebi. Here in this PR, I would like to demonstrate what this kind of update could look like by doing the following:

1. Introduce a typedef appropriate for connecting PSI-MS terms to their allowed value types

   ```
   [Typedef]
   id: has_value_type
   name: has value type
   def: "'Entity A' has value type 'Entity B', such as xsd:float." []
   is_transitive: true
   ```
2. Demonstrate for usages of various datatypes (string, float, double, nonNegativeInteger)

This PR is meant to be a demonstration of what this kind of change would look like and is not complete. It would be easy enough to do some find/replace to fix the rest, but I'm not sure if that follows the typical rules for how OBO terms are laid out.

### DIFF
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -48,6 +48,12 @@ name: has_order
 id: has_domain
 name: has_domain
 
+[Typedef]
+id: has_value_type
+name: has value type
+def: "'Entity A' has value type 'Entity B', such as xsd:float." []
+is_transitive: true
+
 [Term]
 id: MS:0000000
 name: Proteomics Standards Initiative Mass Spectrometry Vocabularies
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -63,8 +63,8 @@ def: "Proteomics Standards Initiative Mass Spectrometry Vocabularies." [PSI:MS]
 id: MS:1000001
 name: sample number
 def: "A reference number relevant to the sample under study." [PSI:MS]
-xref: value-type:xsd\:string "The allowed value-type for this CV term."
 is_a: MS:1000548 ! sample attribute
+relationship: has_value_type xsd:string ! The allowed value-type for this CV term.
 
 [Term]
 id: MS:1000002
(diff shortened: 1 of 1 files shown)


chambm:
I mirror @pporrasebi 's comments about breaking existing tools by changing from xref to has_value_type. However, I don't see why adding has_value_type in addition to the existing xref would break anything. What do you think about having both?

cthoyt:
@chambm that's a great compromise.

I'm currently working on assessing the parsability of all OBO content listed on the OBO Foundry as well as extracting all xrefs. As it is, this xref isn't super meaningful in the context of other xrefs, which mostly constitute equivalence relationships. Additionally, the xref is supposed to point to something like a CURIE, and while this might fit the syntax of a CURIE, it doesn't exactly conform to the semantics of one.

edeutsch:
I agree that there is a lot of tooling that would break if we removed the xref: value-type:
I also agree that relationship: has_value_type looks more elegant and is the way we should have done it way back when
With some reluctance, I agree that having both could perhaps be a useful way for a long transition

cthoyt:
@chambm @edeutsch would you like me to update this PR such that it adds the relationship in all instances? Or do you have a better way to go about doing this?

pabinz:
Fine for me to add a new relationship and to leave the xrefs as is for the moment

edeutsch:
Sure, that sounds good. Please go ahead with a PR. Thanks!

mobiusklein:
@cthoyt I think I made the changes you suggested in #76. Did this improve the parse-ability of the CV? I didn't remove the existing xrefs, but did add the `has_value_type` relationship.

cthoyt:
@mobiusklein besides the recurring typo where there's a slash in `xsd\:string`, this is better. I can close this PR.

PR #53, opened 2021-01-29
replacing alt_id by replaced_by. Closes #44.

### LINKED ISSUE #44: Obsolete term `MS:1000266` should use `replaced_by`, not `alt_id` for replacement
Hi!

Declaring that `MS:1000266` (Laser Desorption) has been replaced by `MS:1000393` is currently not done correctly: the current version of `psi-ms.obo` uses an `alt_id`, which actually assigns the `MS:1000393` identifier to the `MS:1000266`, which is the opposite of what you want to do.

The right way to replace a term is to use the `replaced_by` clause (hopefully OBO-Edit lets you do that correctly):
```ini
[Term]
id: MS:1000266
name: Laser Desorption
def: "OBSOLETE The formation of ions through the interaction of a laser with a material or with gas-phase ions or molecules." [PSI:MS]
comment: This term was made obsolete because it was replaced by laser desorption ionization (MS:1000393).
synonym: "Laser Ionization MERGE" EXACT []
synonym: "LD" EXACT []
is_obsolete: true
replaced_by: MS:1000393

[Term]
id: MS:1000393
name: laser desorption ionization
is_a: MS:1000247 ! desorption ionization
```

sneumann:


edeutsch (approved):
Hi Steffen, apologies for letting this one fall on the floor. We'll get it merged. This seems correct to me.

mobiusklein (approved):

PR #54, opened 2021-01-29
Fix/issue52

sneumann:


mobiusklein:
I've rebased this branch onto the latest `master`, incremented the `data-version` of the vocabulary file and updated the added term's `id` to be unique from the other recently added terms.

Good news is all the tooling worked. Bad news is I still manage to generate commit noise with `git rebase`.

mobiusklein (approved):

edeutsch (approved):
looks good, many thanks!

PR #56, opened 2021-01-29
Fix/issue55

sneumann:
Hi, this PR effectively only contains the few lines:
```
    # Runs a single command using the runners shell
    - name: Check against duplicate IDs
      run: bash -c 'test "X$(grep '^id:' psi-ms.obo | sort | uniq -d)" == "X" '
```
but several commits on the way there. Consider a squash merge 
or pick the above lines manually into validate-obo.yml
Yours, Steffen

### DIFF
--- a/.github/workflows/validate-obo.yml
+++ b/.github/workflows/validate-obo.yml
@@ -30,4 +30,4 @@ jobs:
 
     # Runs a single command using the runners shell
     - name: Check against duplicate IDs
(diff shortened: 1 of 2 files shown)


chambm:
The validator author added a --duplicates check at my request: https://github.com/fastobo/fastobo-validator/issues/39 - I just haven't added it because I wasn't sure when the update would take effect in the odk image. But I just tested and it looks like the --duplicates flag is available now.

sneumann:
Thanks for the info, then this PR is the wrong way. Yours, Steffen

PR #57, opened 2021-02-01
Added new terms for pyteomics and ThermoRawFileParser

caetera:


edeutsch (changes requested):
Just to be a little more complete, would you replace the [PSI-MS] with a PMID and/or a DOI like in other terms like this, e.g.:

[Term]
id: MS:1001969
name: phosphoRS score
def: "phosphoRS score for PTM site location at the PSM-level." [DOI:10.1021/pr200611n, PMID:22073976]
xref: value-type:xsd\:string "The allowed value-type for this CV term."
is_a: MS:1001968 ! PTM localization PSM-level statistic
relationship: has_regexp MS:1002505 ! regular expression for modification localization scoring

edeutsch (approved):
looks great, thanks!

caetera:
Hi Eric,

do you know when the suggestion will be merged into the ontology? It would be nice if we can add it to the release of TRFP that is in preparation now.

chambm (approved):

PR #47, opened 2020-10-19
Adds ProSight and TDPortal terms and mass array

mtrobey:
- adds ProSight and TDPortal terms, along with appropriate parent terms
- adds mass array term for applications such as CDMS that detect neutral masses

PR #60, opened 2021-03-05
fix/issue59

mtrobey:
Fixes a typo in MS:1003130 - the wrong ID was used in the parent term line. This was an error in the original PR #47.

### DIFF
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -1,7 +1,7 @@
 format-version: 1.2
-data-version: 4.1.48
-date: 01:02:2021 19:50
-saved-by: Vladimir Gorshkov
+data-version: 4.1.49
+date: 03:05:2021 11:30
+saved-by: Matthew Robey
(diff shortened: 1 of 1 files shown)


chambm (approved):

chambm:
Hmm, why revert the version? I think even a revision like this should probably bump the final number.

mtrobey:
It looks like the last few issue fix PRs haven't updated the version - let me know if you think I should update it though!

chambm:
@edeutsch ?

edeutsch (changes requested):
yes, please do bump the version number up by one and then we'll merge. Looking at the commit history, it looks like we have been pretty good about this. I don't see a reason to make any change without increasing the version number, especially one that moves a term to a new parent, even fixing an error.

edeutsch:
thanks everyone!

timosachsenberg:
Thanks. I will check and report back.

PR #62, opened 2021-03-24
Add PTMPtophet scores

edeutsch:


chambm (approved):

mobiusklein (approved):
This might be my lack of familiarity with the regex flavors supported, but I thought we only supported ECMAScript syntax, and the patterns these terms reference use POSIX `[:digit:]` character class syntax. Three of the nine regular expressions in the CV use them:
```
regular expression for modification localization scoring MS:1002505`
Regular expression for adduct ion formula MS:1002812
regular expression for interaction scores derived from cross-linking MS:1002665
```
Since this is more to do with connected terms than these, no reason why the answer should block merging regardless.

edeutsch:
> This might be my lack of familiarity with the regex flavors supported, but I thought we only supported ECMAScript syntax, and the patterns these terms reference use POSIX `[:digit:]` character class syntax. Three of the nine regular expressions in the CV use them:
> 
> ```
> regular expression for modification localization scoring MS:1002505`
> Regular expression for adduct ion formula MS:1002812
> regular expression for interaction scores derived from cross-linking MS:1002665
> ```
> 
> Since this is more to do with connected terms than these, no reason why the answer should block merging regardless.

I agree with the last part, and don't know the story on the first part. How about we merge this PR and would you enter in issue regarding the regexps. They are potentially used by validators out there, I don't really know myself. We would need to pull in various people and look at the Java validator for mzIdentML, etc. to understand the implications of the tegexp issues.

For this PR, I was merely following the other similar examples,.

PR #69, opened 2021-05-21
Updated and new terms for ion mobility

edeutsch:
Updated and new ion mobility and other related terms for mzML


mobiusklein (approved):
Looks good to me.

chambm (approved):
Looks fine to me, except I still don't have a way to tell whether Waters data is MSe or not. At least I can tell if it's SONAR though.

PR #71, opened 2021-06-16
add mzQC format to the CV

mwalzer:
We (PSI-QC working group) would like to add mzQC as a file format term in the PSI-MS-CV

### DIFF
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -1,6 +1,6 @@
 format-version: 1.2
-data-version: 4.1.53
-date: 16:06:2021 00:00
+data-version: 4.1.54
+date: 21:06:2021 00:00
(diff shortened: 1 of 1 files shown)


mobiusklein:
Other than the CV version needing to be incremented, nothing here looks amiss.

Do you want to create an intermediate abstract term like "quality control file format" to allow things to refer to QC files of other types?

mwalzer:
> Do you want to create an intermediate abstract term like "quality control file format" to allow things to refer to QC files of other types?
Yep, fair enough.

mobiusklein:
Looks good to me now. Requesting a quick review from at least one other person before merging.

edeutsch (approved):
This looks good to me, thanks!

PR #73, opened 2021-06-17
Add Tryp-N #72

### LINKED ISSUE #72: Add enzyme Tryp-N
Please add enzyme Tryp-N, described in these manuscripts:
* https://www.ncbi.nlm.nih.gov/pmc/articles/PMC7857546/
* https://pubmed.ncbi.nlm.nih.gov/25419962/
* https://www.ncbi.nlm.nih.gov/pmc/articles/PMC7842235/
* https://www.ncbi.nlm.nih.gov/pmc/articles/PMC7995609/

It is also known as LysargiNase, sold by Millipore:
https://www.emdmillipore.com/US/en/product/LysargiNase,MM_NF-EMS0008

Suggested terms to add (instances of `0000000` need to be changed to valid numbers):
```
[Term]
id: MS:0000000
name: Tryp-N
def: "Metalloproteinase found in Methanosarcina acetivorans that cleaves on the N-terminal side of lysine and arginine residues." [https://www.ncbi.nlm.nih.gov/pmc/articles/PMC7857546/]
is_a: MS:1001045 ! cleavage agent name
relationship: has_regexp MS:0000000 ! (?=[KR])

[Term]
id: MS:0000000
name: (?=[KR])
def: "Regular expression for Tryp-N." [PSI:PI]
is_a: MS:1001180 ! Cleavage agent regular expression
```

mobiusklein:
Closes #72 

Also added a handy script to find gaps in accession number space.

### DIFF
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -1,7 +1,7 @@
 format-version: 1.2
-data-version: 4.1.53
+data-version: 4.1.54
 date: 21:05:2021 00:00
-saved-by: Eric Deutsch
+saved-by: Joshua Klein
 auto-generated-by: OBO-Edit 2.3.1
 import: http://ontologies.berkeleybop.org/pato.obo
 import: http://ontologies.berkeleybop.org/uo.obo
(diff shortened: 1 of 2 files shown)


edeutsch (approved):
I think our custom is to specify the PMID instead of the PMCentral URL, but not a big deal.
PMID:33517145

edeutsch (changes requested):
Pondering on it further, I think this is the appropriate citation: 25419962, not the one in the PR which is not the original paper, just a recent use.
Also, I think it is better known as LysargiNase and is the name used in the original paper, so I think at least an exact synonym, if not the primary name, of LysargiNase is useful.
I'm not certain of this, so a second look would be useful.

mobiusklein:
Updated to use LysargiNase as the name with a synonym entry added. I agree, I found more hits on Pubmed searching LysargiNase than Tryp-N.

edeutsch (changes requested):
Thanks for the updates. Sorry, one more quibble, would you set the citation to PMID:25419962 unless you disagree that that is a better citation?

mobiusklein:
I thought I had taken your specified citation, sorry for the confusion.

edeutsch (approved):
great, many thanks!

edeutsch:
Maybe @chambm or @pabinz could do a quick double check to make sure this is all appropriate and approve?

chambm (approved):

Issue #51, opened 2021-01-28
Document Term contribution process


sneumann:
Title says it all. Maybe README is sufficient as a start. 
Do we want to have 1) An issue template for new terms, 
and/or simply ask for pull requests ? Add a remark where IDs come from 
(in case of 2) self chosen ?).
My main difficulty is that we can't have N simultaneously open PR each appending one Term at the end, and we can't have one big PR adding N terms if we want to be able to reject just one of the terms. 
Yours, Steffen

mobiusklein:
Is the problem:

1. That it's unclear how to initiate a new term request
2. Because a PR with a formatted term will inevitably have outdated IDs as competing requests all pick overlapping sequential ID values
3. Because multiplexing terms in a PR leads to undesirable gaps in ID sequences

Issue 1 would be solved with an issue template, easily enough. Issues 2 and/or 3 would require more tooling. I'm partial to the idea of using a pattern like `MS:100$X` within the PR's text and then we run a script on `psi-ms.obo` that finds the next available ID that is a suffix of `MS:100` (and does this for sequentially all such variables in the file). This would definitely require an issue template to explain too though.

sneumann:
Hi, basically it is 1. , so a README plus issue template is a good start.
I wouldn't overengineer for 2. and 3. if multiple Terms are submitted, 
their IDs  need to be coordinated, so the maintainer accepting the PR may have 
to edit the PR. Yours, Steffen

edeutsch:
@sneumann okay to close?

sneumann:
Yes, README is fine. The issue templates are a bit oversimplified, as we're not asking for where in the tree. But that might actually be good for new contributors and be figured out with a few messages forth and back. 
Thanks, Yours, Steffen

PR #74, opened 2021-06-24
Add README text and issue template for new term request process

mobiusklein:
This is a first pass at addressing #51 in the README.

I also tried adding an issue template, but it seems like GitHub has changed the format for these recently.


cbielow:
this would have helped me in search of `how to add my CV term` :)
Can this be merged?

edeutsch (approved):
great, thanks!

PR #75, opened 2021-06-25
add PTX-QC as software

cbielow:
Hi all,

this PR adds PTX-QC as analysis software (to enable referencing itself when writing mzQC files).

I also added https://github.com/HUPO-PSI/psi-ms-CV/commit/5accf6c2de9364a5a1bd3b905b4e48c5651808ae, because OBO-Edit software would not allow me to save the obo file, because:
```
1 fatal error:
PTMProphet normalized information content (MS:1003149) generated 1 error:
  Definition of MS:1003149 cannot contain extended characters.
```

which apparently points to the hyphen `–`, which is (according to OBOEdit) not allowed.
I can revert that part though if desired.

### DIFF
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -20561,7 +20561,7 @@ is_a: MS:1001459 ! file format
 [Term]
 id: MS:1003162
 name: PTX-QC
-def: "Proteomics (PTX) - QualityControl (QC) software for QC report generation and visualization" [DOI:10.1021/acs.jproteome.5b00780, PMID:26653327, https://github.com/cbielow/PTXQC/]
+def: "Proteomics (PTX) - QualityControl (QC) software for QC report generation and visualization." [DOI:10.1021/acs.jproteome.5b00780, PMID:26653327, https://github.com/cbielow/PTXQC/]
 is_a: MS:1001456 ! analysis software
 synonym: "PTXQC" EXACT []
 
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -1,7 +1,7 @@
 format-version: 1.2
-data-version: 4.1.55
-date: 21:05:2021 00:00
-saved-by: Joshua Klei
+data-version: 4.1.56
+date: 25:06:2021 00:00
+saved-by: Chris Bielow
 auto-generated-by: OBO-Edit 2.3.1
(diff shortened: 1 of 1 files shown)


mobiusklein:
The character problem is because the `–` is an "en-dash", not a hyphen, so it's not ASCII. Shouldn't be an issue.

Is "quality control software" a category distinct from "analysis software"? Otherwise, this seems fine to me.

psi-ms.obo
```diff
@@ -20557,6 +20558,13 @@ name: quality control data format
 def: "Grouping term for quality control data formats." [PSI:MS]
 is_a: MS:1001459 ! file format
 
+[Term]
+id: MS:1003162
+name: PTX-QC
+def: "Proteomics (PTX) - QualityControl (QC) software for QC report generation and visualization" [DOI:10.1021/acs.jproteome.5b00780, PMID:26653327, https://github.com/cbielow/PTXQC/]
+is_a: MS:1001456 ! analysis software
```
mobiusklein, line 20565: Should this be a new sub-type of `software` "quality control software" instead of "analysis software"?

cbielow: its kind of both, so I'm happy either way

chambm: If we don't already have a QC software category (and I'm not convinced we need it), analysis software seems fine. QC software would be a child of analysis software IMO.

chambm (approved):

edeutsch (approved):
This all looks great to me, except I think there is a formal OBO requirement that definitions end in a period? I thought OBO-Edit enforced this.

cbielow:
> 
> This all looks great to me, except I think there is a formal OBO requirement that definitions end in a period? I thought OBO-Edit enforced this.

True indeed. OboEdit says:
> PTX-QC (MS:1003162) generated 1 warning:
> Definition of MS:1003162 does not end with a period, question mark or exclamation point.

I added the dot. Got overlooked because its buried in another 776 non-critical warnings when storing the file.

(I did not use push the .obo generated by OBOEdit2 here, because it changes a whole lot of other things, which are not part of the PR and would just have added noise.)

mobiusklein (approved):

PR #76, opened 2021-06-26
Add has_value_type

mobiusklein:
Implements the same concept as #50 formally across the entire controlled vocabulary.

I also added three formal data type terms:

```
[Term]
id: MS:1002709
name: compound data type
def: "A data type representing more than a single value." []

[Term]
id: MS:1002710
name: list of type
def: "A data type defining a list of values of a single type." []
is_a: MS:1002709 ! compound data type

[Term]
id: MS:1002711
name: list of strings
def: "A list of xsd:string." []
is_a: MS:1002710 ! list of type
relationship: has_value_type: xsd\:string

[Term]
id: MS:1002712
name: list of integers
def: "A list of xsd:integer." []
is_a: MS:1002710 ! list of type
relationship: has_value_type: xsd\:integer

[Term]
id: MS:1002713
name: list of floats
def: "A list of xsd:float." []
is_a: MS:1002710 ! list of type
relationship: has_value_type: xsd\:float
```
and an implementation for mzSpecLib:
```
[Term]
id: MS:1003162
name: analyte mixture members
def: "The set of analyte identifiers that compose an interpretation of a spectrum." [PSI:PI]
is_a: MS:1003078 ! interpreted spectrum attribute
relationship: has_value_type MS:1002711 ! list of strings
```

### DIFF
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -1,5 +1,5 @@
 format-version: 1.2
-data-version: 4.1.56
+data-version: 4.1.57
 date: 21:05:2021 00:00
 saved-by: Joshua Klein
 auto-generated-by: OBO-Edit 2.3.1
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -17799,21 +17799,18 @@ id: MS:1002711
 name: list of strings
 def: "A list of xsd:string." []
 is_a: MS:1002710 ! list of type
-relationship: has_value_type xsd\:string ! None
 
 [Term]
 id: MS:1002712
 name: list of integers
 def: "A list of xsd:integer." []
 is_a: MS:1002710 ! list of type
-relationship: has_value_type xsd\:integer ! None
 
 [Term]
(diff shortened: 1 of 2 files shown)


chambm (changes requested):
I see a lot of changes here other than adding has_value. Why were comments and is_obsolete keys removed?

edeutsch (commented):
I agree, all those other lines should be preserved. I suppose the script that autogenerated the 'relationship' changes needs to be updated to preserve other lines in the CV. Thanks!

mobiusklein:
I've reversed the addition of all the new value-type terms, but fixed the lost information during the first automated attribute addition.

edeutsch (approved):
This all looks great to me, thanks! The only possible flaw is that the diff seems to show that term MS:1003162 is being deleted? Shouldn't that term stay? Is that a bug or intentional?

edeutsch (approved):
all good now, thanks!

PR #78, opened 2021-08-11
Add QuaMeter and iMonDB software

bittremieux:
Similar to #75, additional tools for QC metric calculation.

### DIFF
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -1,5 +1,5 @@
 format-version: 1.2
-data-version: 4.1.57
+data-version: 4.1.58
 date: 11:08:2021 00:00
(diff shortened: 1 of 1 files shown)


edeutsch (approved):

bittremieux:
QuaMeter IDFree is pretty much the default version we use currently. The original version implementing the NIST QC metrics still exists, but we haven't been using it ourselves recently because it's more cumbersome to use and doesn't really provide more relevant data.

mzQC can also support QuaMeter IDbased, and a bunch of other QC tools as well, it doesn't restrict which tools can be used. It's just that for our current examples we haven't needed CV terms for those yet.

PR #84, opened 2021-09-10
Alternative EThcD with new parent term

edeutsch:


mobiusklein (approved):
As discussed, the supplemental energy term resolves my concerns about ambiguity

ypriverol (approved):
@edeutsch I see that we have changed the names of all therms to lowercase for the first letter. Is that a consensus. ? 

@mobiusklein @edeutsch the check for obo  is failing.

mobiusklein (approved):
Updated header

edeutsch:
> @edeutsch I see that we have changed the names of all therms to lowercase for the first letter. Is that a consensus. ?

I believe that this is our custom, and thus I made the change to conform to the custom.
I don't think this term is currently used by ProteoWizard so the likelihood of breaking existing code seems low. but possible.
I think it's a good idea. But don't feel strongly.
If anyone objects, please speak up!

chambm:
Yes, pwiz is not using these terms. I missed the conversation about why we need combined terms to reduce ambiguity vs. the old approach of having both the CID/HCD term and the ETD term. Is the new way supposed to be in addition to or instead of the old way?

edeutsch:
My suggestion is for pwiz not to do anything differently and the recommendation for mzML is to keep doing what we've always been doing. But for other applications that want a single term, it seems useful to have these two combined modes. 

Note that there is a term "supplemental collision energy" to capture the collision energy of the supplemental activation for these combined modes.

chambm:
Seems pretty confusing to handle it different between different formats, but I guess that's not new. :/

chambm (approved):

PR #88, opened 2021-10-17
Automatically generate OWL from OBO, tag, and release

mobiusklein:
This PR adds an automated operation whenever a commit is pushed to `master` that will generate an OWL file from the OBO, check if the OWL file changed, and if it did change, generate a new tag and release.


mobiusklein:
Rebased to remove the pile of garbage commits used to debug the action script (and obscure my many typos). 

Some implementation notes:

1. While ontodev/robot does have a Dockerfile it does not have a published Docker container. I've opened an issue with them to see if this can be rectified but in the mean time I am just downloading the jar file from their releases page. This does not interfere with the automatic generation of the OWL file or the extra commits.
2. I was able to test the automated tagging without making  a mess, but was unable to get the subsequent release generation to work (the tag was already created).
3. When a tag already exists (or the OBO version doesn't change) the tagging process aborts and this workflow logs as a failure. This probably not ideal in the long run but it doesn't seem to be possible to "abort" without an error.
4. This workflow only runs when the OBO file changes, so if we eventually need to version based upon multiple files, this will need to be reworked.

This action currently uses an access token generated from my account with a limited lifespan. We'll see if the process works well enough and if so I'll make a service account for more general HUPO-PSI automation operations.

mobiusklein:
I was directed from https://github.com/ontodev/robot/issues/917 to the official Docker container for `robot` being published elsewhere. I was able to switch to using that instead of downloading the jar file on each run, though arguably downloading a jar file is less work than downloading a VM image containing the jar file.

PR #92, opened 2021-11-24
Add missing mzSpecLib terms

mobiusklein:


psi-ms.obo
```diff
@@ -21985,6 +21985,53 @@ is_a: MS:1003201 ! library provenance attribute
 xref: value-type:xsd:string "The allowed value-type for this CV term."
 relationship: has_value_type xsd\:string ! The allowed value-type for this CV term
 
+[Term]
+id: MS:1003207
+name: library creation software
+def: "Library creation software" [PSI:MS]
+is_a: MS:1000531 ! software
+is_a: MS:1003171 ! spectral library attribute
+relationship: has_value_type xsd\:string ! The allowed value-type for this CV term
```
chambm, line 21994: Are we not encouraging software vendors to add their software as CV terms and using those whenever possible instead of free text?

psi-ms.obo
```diff
@@ -21985,6 +21985,53 @@ is_a: MS:1003201 ! library provenance attribute
...
+
+[Term]
+id: MS:1003208
+name: reported precursor monoisotopic m/z
+def: "The measured or inferred (as reported by the mass spectrometer acquisition software or post-processing software) of the monoisotopic peak of the precursor ion based on the MSn-1 spectrum." [PSI:MS]
```
chambm, line 21999: Isn't this redundant with `selection ion m/z`?

psi-ms.obo
mobiusklein: I had that question too, but the entry from the CV terms to create table had included this value type expectation.

I think registration is preferable to free text for consumers of CV term containing documents, but for new tool writers it is a chore.

chambm: Yes it's a chore but it's also the status quo. It's going to make me twitch if some of the PSI formats only support free text and others don't. :stuck_out_tongue:

psi-ms.obo
mobiusklein: This is specifically about the *monoisotopic m/z*, not whatever peak the acquisition software chooses to report (e.g. the most abundant peak or the one in the center of the isolation window). I've seen  Thermo products will report different peaks on different instrument models or with different instrument method settings, or depending upon whether or not certain bugs remain unpatched.

chambm: The most abundant peak, often in the center of the isolation window, will usually be the isolation target m/z. How/where do you see this term being used? How do you tell whether the reported precursor m/z is monoisotopic or not without doing your own processing, which makes it no longer really the reported value? Or are you proposing to report whatever the instrument says without checking it? That would seem equivalent to selected ion m/z (which, at least by convention, will be the instrument's reported "monoisotopic" m/z if available and the isolation target m/z if not).

mobiusklein: This term is primarily intended for use in spectral libraries where we want to express explicitly that this is what the _library creation software_ thinks is the monoisotopic peak, how ever it is convinced. If I were making a library, I'd take the monoisotopic m/z from the output of my charge state deconvolution and deisotoping algorithm as this value, maybe averaged over time from the linked MS1 evidence. I may write that value back out to mzML, though I would still use the old term there because that's what the reader is expecting.

psi-ms.obo
mobiusklein: I think I remember the issue now. With mzSpecLib, the design allows you to have one CV param whose value is another CV param. This is because mzSpecLib is structured almost entirely by CV params instead of a formal schema breaking down the data model.

In XML, we'd write:
```xml
<LibraryCreationSoftware version="1.2.3" id="creation_software1">
    <cvParam accession="MS:1001477"  name="SpectraST" />
</LibraryCreationSoftware>
``` 
In mzSpecLib, that could be constructed like this:
```ps1
MS:1003207|library creation software=SpectraST
```
or like this (redundantly?):
```ps1
MS:1003207|library creation software=MS:1001477|SpectraST
```
Here though the value is a term, not a string, and we haven't formally stipulated that situation yet.

However, if we weren't insisting on the key naming the scope and just hope we can tell by context as we can in the narrowed context in an XML file, we could just write:
```ps1
MS:1001477|SpectraST
```
and then the reader would have to deduce:
1. The current term names library creation software (amongst other things)
2. The current scope can accept library creation software
3. This term is therefore _the_ library creation software

This is not ideal if there's ambiguity. In this case it would be pretty contrived, but I'm not sure we can prove that universally.

chambm: If the value is a CV term/accession, I think that should somehow be specified instead of simply allowing any string. But there also needs to be a way to denote that the term can't be used that way in mzML (e..g `<cvParam accession="MS:1003207" name="library creation software" value="MS:100147|SpectraST" />`). The only current way I can think to do that is to make sure the term doesn't map into mzML, by not making it a child of software. Which seems like a pretty yucky solution.

psi-ms.obo
chambm: In that case I'd drop "reported" from the name because that strongly implies coming from the acquisition software rather than post-processing. Perhaps "experimental" would be a better modifier? That's what mzIdentML uses (as an XML attribute).

edeutsch: I think @chambm has a good point here. I think we wanted a word to differentiate from theoretical. But mzIdentML uses SpectrumIdentificationItem.experimentalMassToCharge:
```
experimentalMassToCharge  xsd:double required  The mass-to-charge value measured in 
the experiment in Daltons / charge. 
```
Curiously mzIdentML doesn't try to stipulate monoisotopic, but I suppose that is implied?
I think "experimental precursor monoisotopic m/z" is an improvement.

psi-ms.obo
edeutsch: I agree that we made a mistake by making this a free text value. We should go with a term without a value. That did not occur to me at the time. So I think we want to see this:
```
MS:1003207|library creation software=MS:1001477|SpectraST
```
in the format. so the revised entry should be this:
```
[Term]
id: MS:1003207
name: library creation software
def: "Library creation software" [PSI:MS]
is_a: MS:1000531 ! software
is_a: MS:1003171 ! spectral library attribute
```
i.e. no relationship: has_value_type.
and we should make sure that SpectraST and all other software terms that can create libraries are children of this term (not necessarily exclusively)

Does that seem agreeable?

psi-ms.obo
mobiusklein: I can see a convoluted way to argue that anything other than arbitrary assignment of a random value is "experimental" to a point when you use post-acquisition software to adjust the precursor monoisotopic peak assignment based on any observation from that data. I can make that change.

I don't think mzIdentML really cares about monoisotopic masses outside of modifications. That's the only time the word "monoisotopic" appears in the spec.

edeutsch: It seems odd to me that "monoisotopic" should be called out specifically in the context of mass modifications but not in the context of precursors. I assumed that this was just accidental rather than intentional.

psi-ms.obo
mobiusklein: It doesn't express that the term can take a value, so one might not expect `MS:1003207|library creation software=MS:1001477|SpectraST` to be legal.

Another option, purely expedient and paying no heed to good form is to add a provision to the mzSpecLib specification that lets you use `<term1>=<term2>` statements where `term2` is an instance of `term1` to stipulate you are using `term2` in the role of `term1` explicitly. I am thinking of this like type casting in typed languages (`static_cast<char*>(stringvar)`, `(Base)Derived`). This gets us around that ambiguity problem, and leaves us free of recursive value type specifications.

edeutsch: Isn't that what we are already doing? We already have all these in our examples:
```
MS:xxxxxxx|instrument model=MS:1000449|LTQ Orbitrap
MS:xxxxxxx|analyte type=MS:xxxxxxx|peptide
MS:1000525|spectrum representation=MS:1000127|centroid spectrum
MS:1000034|intensity unit=MS:1000905|percent of base peak times 100
MS:1000901|retention time normalization standard=MS:1002005|iRT retention normalization standard
MS:1000465|scan polarity=MS:1000130|positive scan
```

mobiusklein: You're right, I hadn't seen that. Removed the `has_value_type` clause.

psi-ms.obo
mobiusklein: Renamed to "experimental precursor monoisotopic m/z"

chambm (approved):

edeutsch (approved):

edeutsch (approved):
oh, I suppose the date should be updated, but then ready to merge! Many thanks!

PR #95, opened 2021-12-17
Resurrect old obsolete term for new 'release date'

edeutsch:
This term was obsoleted eons ago in favor of an XML attribute. Here we resurrect the term and repurpose it for a slightly more generic usage. Note that version information at the top of the file must still be updated in PR when it comes time to merge.


edeutsch:
Leave this unmerged pending pending discussion in #94

PR #96, opened 2022-01-05
Switch to PURL links for OBO imports

chambm:
@cmungall is this what you mean?


psi-ms.obo
```diff
@@ -1,10 +1,10 @@
 format-version: 1.2
-data-version: 4.1.65
+data-version: 4.1.66
 date: 04:12:2021 00:00
 saved-by: Eric Deutsch
 auto-generated-by: OBO-Edit 2.3.1
-import: http://ontologies.berkeleybop.org/pato.obo
-import: http://ontologies.berkeleybop.org/uo.obo
+import: http://purl.obolibrary.org/obo/pato.obo
```
cmungall: Much better!

Even better would be to use the OWL PURL. As you are hand editing the obo file I don't think this will negatively impact you.

psi-ms.obo
```diff
@@ -16064,6572 +16064,4 @@ relationship: has_value_type xsd\:double ! The allowed value-type for this CV te
...
 name: distinct peptide-level FDRScore
-def: "MzidLibrary FDRScore for distinct peptides once redundant identifications of the same peptide have been removed (id est multiple PSMs have been collapsed to one entry)." [PSI:PI]
-xref: value-type:xsd\:double "The allowed value-type for this CV term."
-is_a: MS:1002358 ! search engine specific peptide sequence-level identification statistic
-is_a: MS:1002363 ! search engine specific score for proteins
-relationship: has_domain MS:1002349 ! value greater than zero but less than or equal to one
-relationship: has_value_type xsd\:double ! The allowed value-type for this CV term
-
-[Term]
-id: MS:1002361
```
cmungall: I assume this is a mistake. Terms can be obsoleted but they should never be deleted. Same for ones below

cmungall (changes requested):

psi-ms.obo
chambm: Yeah the GitHub editor likes to truncate large files but it only happens on commits after the first one for some reason.

psi-ms.obo
chambm: Would that mess up people using OBO-Edit?

psi-ms.obo
edeutsch: So what's the solution? Will you start a new PR? or repair this one?

edeutsch (changes requested):
This PR deletes many lines accidentally.

psi-ms.obo
chambm: I'll start a new PR when I get an answer about whether changing to OWL imports messes up OBO-Edit.

edeutsch: okay. I don't really understand the deep implications here, but I am thinking that in our OBO file, the reference should stay as OBOs (as you have done). Then, @mobiusklein already has the magic built in (although maybe needs tweaking with this change?) to transform the OBO references to OWL references in the translated OWL. So, our OBO references other OBOs, and our OWL references other OWLs. That seems best to me, but my understanding is limited. Fine to wait for an answer from @cmungall

cmungall: Ah, I wasn't aware you were still using obo-edit, I thought you were solely hand-editing. You have various options. Before exploring these, can I ask, why are the imports even in the file? The main reason to do an import of an external ontology is to link to those terms from within your ontology. But you aren't even using UO or PATO in MS. What is the use case for these imports? Is it to ensure that these terms are shown alongside MS in browsers like MS? Wouldn't you want to extract a smaller subset here? (I have a general feeling there are a number of historic artefacts in your file and you could make things simpler for yourself by revisiting some of these).

If you really wish to keep these imports and you wish to keep using obo-edit, you have two options:

1. use the .owl purls, and include an obo-edit catalog called `oboedit.catalog` that is a 2 column file mapping the logical names to paths on disk
2. use the .obo purls

I think 1 is better, but entails more work (made harder by the fact oboedit has been unsupported for almost a decade)

Also, I don't know if you are aware, but you can edit .obo files in Protege, it will stay saved as .obo. Many projects like GO, Uberon maintain the source for their ontologies as .obo and edit in Protege (retaining the benefits of easy manual edits and clear diffs).

cmungall: here is an example of an oboedit catalog https://github.com/geneontology/go-ontology/blob/master/src/ontology/oboedit.catalog (this one is no longer in use, as we edit the obo file exclusively using Protege in GO now)

edeutsch: We are using UO like this in MS:
```
relationship: has_units UO:0000098 ! milliliter
..
relationship: has_units UO:0000175 ! gram per liter
```

so it seems like we should keep it for that reason?

But you are right that there don't seem to be references to PATO anywhere. So maybe we could just remove that one. Any past reason for its presence presumably has been removed.

edeutsch: > Also, I don't know if you are aware, but you can edit .obo files in Protege, it will stay saved as .obo. Many projects like GO, Uberon maintain the source for their ontologies as .obo and edit in Protege (retaining the benefits of easy manual edits and clear diffs).

Thanks. I recall hearing about this before, installing Protege, trying it, and not getting it to work and giving up. I probably should have asked for assistance before giving up.

PR #97, opened 2022-01-05
Switch to PURL for OBO imports

chambm:
@edeutsch is right that the robot convert step is already changing the OBO imports to OWL.

### DIFF
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -1,10 +1,10 @@
 format-version: 1.2
-data-version: 4.1.65
+data-version: 4.1.66
 date: 04:12:2021 00:00
 saved-by: Eric Deutsch
 auto-generated-by: OBO-Edit 2.3.1
-import: http://ontologies.berkeleybop.org/pato.obo
-import: http://ontologies.berkeleybop.org/uo.obo
+import: http://purl.obolibrary.org/obo/pato.obo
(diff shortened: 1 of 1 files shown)


mobiusklein:
Yes, https://github.com/HUPO-PSI/psi-ms-CV/blob/master/.github/workflows/tag-pr.yml#L29-L45 takes care of that conversion. By changing the URL in the OBO, it will also be reflected in the OWL file.

edeutsch:
great, it looks like it is location independent, so if we change the URL a bit, it should still work fine. thanks!

chambm:
Hmm any idea why it failed @mobiusklein ? https://github.com/HUPO-PSI/psi-ms-CV/runs/4729786428?check_suite_focus=true

mobiusklein:
It seems `robot` couldn't handle an `import` of OBO files from the new host. If I switch the extensions on the URIs to import it works without issue. That's a simple enough hacky fix. Looking at the HTTP headers, the only significant difference between the old host and the new one is that the new one uses at least one HTTP redirect while the other does not.

I'll add the work-around.

mobiusklein:
It seems `robot` couldn't handle an `import` of an OBO, as it never tries an OBO parser on the import, trying every other thing under the sun.

mobiusklein:
Fixed in #98

edeutsch:
great, thank you for fixing!

Issue #80, opened 2021-08-18
QC-CV -> MS-CV


mwalzer:
### Describe the question or discussion

Dear PSI-MS CV community,  
_( I posted this message to the PSI-MS-CV mailing list, but is stuck in the approval queue. Sorry for the duplication, GitHub issues is a better place to have such discussions in any case )_

I am writing on behalf of the PSI-QC working group, as we want to get our QC-CV 'into the fold'. Before that, however, we thought it'd be best to start a conversation on how. 
We have been developing a separate CV initially to have some freedoms on what/how things work out best for the mzQC format, which relies heavily on CV control. 
Our CV holds the metric definitions, relations and few additional 'organisational' terms, so far <300 entries, for which we chose the QC prefix (naturally). We ditched the `xref: value-type` declarations in favour of defining term values with full CV terms and relations, which works nicely with the format. 
The mzQC format is, unlike most PSI formats, JSON based. One of the goals in the design was to bring the format closer to the last miles of analysis (R/pandas/web) but also as something that works between any steps of an MS-based study workflow (lightweight with high automation potential). If you want to know more, check out our specification document draft ( https://hupo-psi.github.io/mzQC/). For mzQC, we have to cover a vast range of QC/MS-related topics in the CV. For that, we've devised a set of templates and rules for adding new term/metrics via GitHub issues. To keep better track of the entries, Nils created a GitHub workflow for obo-to-markup conversion to enhance browsability. QC-CV is not standalone, we base as much as possible on existing definitions, so we imported few other CVs: UO, OBI, STATO (,NCIT) and PSI-MS of course. So, overall I think QC-CV would fit in nicely, either as big PR or piece-wise. As for the merge, maybe we can follow the same playbook as the PEFF integration did? What do you think? 

**TL;DR**: mzQC folks want to merge their CV into PSI-MS. But before opening a gargantuan PR, we should talk. 

Best,

edeutsch:
Hi Mathias, thanks for contacting us. Your message did actually go out to the list quite rapidly, but you didn't get a notification that it went out, just that it was briefly held. so everyone has seen it. We're just slow.

We should definitely set up a conversation about this. A few comments from me:
- Where can we have a look at what you have so far?
- Shall we set up a doodle poll to select a time when interested parties can meet?
- What is your level of urgency? It is high holiday time now
[...]

PR #99, opened 2022-01-27
Added first batch of QC terms for mzQC in PSI-MS-CV

mwalzer:
This is the first batch of QC related terms in use with mzQC. As discussed in #80 we'd like to integrate into PSI-MS-CV and changed our mzQC terms as much as possible to align with the way terms are represented in PSI-MS-CV.
The first batch of terms is for one a rough cross-section of how we designed terms and concepts and also the extent to which we integrated QuaMeter metrics. 
As also discussed in #80 and the subsequent telco, we added the terms in a separate section of the namespace-id `MS:$sequence(7,4000000,4999999)$`. We then pinned our terms to the PSI-MS-CV root as a side-branch with:
```
[Term]
id: MS:4000000
name: PSI-MS CV Quality Control Vocabulary
def: "PSI Quality Control controlled vocabulary term." [PSI:QC]
relationship: part_of MS:0000000 ! Proteomics Standards Initiative Mass Spectrometry Vocabularies
```

Other notable introductions:
We added Typedef `has_value_concept`  where `has value type` is not enough; e.g.: where a float value represents a ratio. 
We added Typedefs `has_column` and `has_optional_column` to reflect the table value type as explained in #80 and the telco.
We introduced imports for the units and concepts, namely unit ontology, the statistical methods ontology, and the NCI thesaurus ontology.

You can find examples of how we use the terms in mzQC for example [here](https://github.com/HUPO-PSI/mzQC/blob/main/doc/examples/QC2-sample-example.mzQC).


edeutsch:
For public visibility, we discussed the proposed terms last Friday and came up with some suggestions on how to improve them. Here is the sheet we worked through: https://docs.google.com/spreadsheets/d/15v694KU_1Fx7cmCEtgynIaqUpK13feCIA0xaemZoNF4/edit#gid=0
Wout was going to discuss these with the mzQC team to see if they can be accommodated.

mwalzer:
Thank you @edeutsch and the PSI-MS-CV group for the swift consideration of the PR!
We discussed the suggestions and incorporated all but one, *not* changing the single value type name to scalar.
We hope the changes are sufficient.

psi-ms.obo
```diff
@@ -20,12 +21,20 @@ remark: creator: Fredrik Levander <fredrik.levander <-at-> immun.lth.se>
...
+remark: creator: Nils Hoffmann < nils.hoffmann <-at-> cebitec.uni-bielefeld.de>
+remark: creator: Julian Uszkoreit <julian.uszkoreit <-at-> ruhr-uni-bochum.de>
+remark: creator: Mathias Walzer <walzer <-at-> ebi.ac.uk>
 remark: publisher: HUPO Proteomics Standards Initiative Mass Spectrometry Standards Working Group and HUPO Proteomics Standards Initiative Proteomics Informatics Working Group
 remark: When appropriate the definition and synonyms of a term are reported exactly as in the chapter 12 of IUPAC orange book. See http://www.iupac.org/projects/2003/2003-056-2-500.html and http://mass-spec.lsu.edu/msterms/index.php/Main_Page
 remark: For any queries contact psidev-ms-vocab@lists.sourceforge.net
 remark: URL: https://raw.githubusercontent.com/HUPO-PSI/psi-ms-CV/master/psi-ms.obo
 remark: This work is licensed under the Creative Commons Attribution 4.0 International (CC BY 4.0) license.
 remark: To view a copy of this license, visit https://creativecommons.org/licenses/by/4.0/ or send a letter to Creative Commons, 444 Castro Street, Suite 900, Mountain View, California, 94041, USA.
+import: http://ontologies.berkeleybop.org/uo.obo
```
mobiusklein, line 35: This line is redundant with line 7.

psi-ms.obo
```diff
@@ -59,6 +68,18 @@ id: has_structured_representation_in_format
...
+name: has_value_concept
+
+[Typedef]
+id: has_column
+name: has_column
+
+[Typedef]
+id: has_optional_column
+name: has_optional_column
+
```
mobiusklein, line 82: The typedef for `has_metric_category` is missing

mobiusklein (changes requested):
Beyond the two notes, everything else looks good.

psi-ms.obo
```diff
@@ -20,12 +21,20 @@ remark: creator: Fredrik Levander <fredrik.levander <-at-> immun.lth.se>
...
+remark: creator: Mathias Walzer <walzer <-at-> ebi.ac.uk>
 remark: publisher: HUPO Proteomics Standards Initiative Mass Spectrometry Standards Working Group and HUPO Proteomics Standards Initiative Proteomics Informatics Working Group
 remark: When appropriate the definition and synonyms of a term are reported exactly as in the chapter 12 of IUPAC orange book. See http://www.iupac.org/projects/2003/2003-056-2-500.html and http://mass-spec.lsu.edu/msterms/index.php/Main_Page
 remark: For any queries contact psidev-ms-vocab@lists.sourceforge.net
 remark: URL: https://raw.githubusercontent.com/HUPO-PSI/psi-ms-CV/master/psi-ms.obo
 remark: This work is licensed under the Creative Commons Attribution 4.0 International (CC BY 4.0) license.
 remark: To view a copy of this license, visit https://creativecommons.org/licenses/by/4.0/ or send a letter to Creative Commons, 444 Castro Street, Suite 900, Mountain View, California, 94041, USA.
+import: http://ontologies.berkeleybop.org/uo.obo
+import: http://purl.obolibrary.org/obo/stato.owl
+import: http://purl.obolibrary.org/obo/ncit.owl
```
edeutsch, line 37: should we reference the obos here in the obo?
And then transform to owl only in the transformed owl?

psi-ms.obo
mwalzer: indeed, thanks. Should we move line36-37 after line7?

psi-ms.obo
mwalzer: added in b7f9ca66a11fd8bbe779e4ec665d538529a84873

psi-ms.obo
mwalzer: OBOfoundry, OLS, stato github, etc. only offer the owl version. 
NCIT is here (http://purl.obolibrary.org/obo/ncit.obo). However, not directly on the page with the purl link but on some other part of obofoundry it states for NCIT: `NCIt OBO Edition releases should be considered experimental.`
Unsure how to proceed.

edeutsch: okay, that's fine, we'll leave as is.

edeutsch (approved):

psi-ms.obo
mobiusklein: For simplicity, that seems reasonable.

mobiusklein (changes requested):

mobiusklein (approved):

PR #100, opened 2022-01-28
add paper spray ionization term

edeutsch:


chambm (approved):
I have no idea what that is but the definition looks sane. :+1:

edeutsch:
Thanks. I probably should have provided a little more context:

Initiating request:
https://github.com/compomics/ThermoRawFileParser/issues/128

Wikipedia entry that provided some inspiration:
https://en.wikipedia.org/wiki/Paper_spray_ionization

mobiusklein (approved):
Looks good to me

PR #104, opened 2022-02-12
mzSpecLib Terms Batch 4

mobiusklein:
This batch of terms adds several terms missing from the spreadsheet, and covers the following renamings:
1.  `spectrum position` -> `library entry key` and `spectrum index` -> `library entry index`
2. `library entry class` -> `library spectrum attribute set`


edeutsch (commented):
I'm not certain I'm reading this right, but I think this PR stomps on the previous commit for paper spray ionization MS:1003235 (which I may have misplaced after the MS:4000000 series?)? I'm also wondering why it skips from 71 to 73?

mobiusklein:
I re-arranged terms to be in order of ID, so `paper spray ionization` got moved to before the mzQC terms, which are all still intact.

I can't seem to install `fastobo-validator` from source or binary so I am resorting to using GH actions to find the syntax error.

edeutsch (commented):
oh, I see it now, okay. But it's identifier changed from 3235 to 3236. Is that necessary? ThermoeRawFileParser would seem to be using 3235 already:

https://github.com/compomics/ThermoRawFileParser/pull/129/commits/ee973e8ed96254a6b39bd4b102f14da0a993a09b

mobiusklein:
There was a collision from earlier work on the branch, I re-arranged the terms so the original ID is restored for `paper spray ionization`.

edeutsch (approved):
many thanks, this all looks good to me!

psi-ms.obo
```diff
@@ -22196,6 +22198,68 @@ is_a: MS:1001456 ! analysis software
 is_a: MS:1003207 ! library creation software
 is_a: MS:1002878 ! small molecule analysis software
 
+[Term]
+id: MS:1003234
+name: library entry attribute
+def: "An attribute that describes information about an entry in a sepctral library" [PSI:PI]
```
henryhlam, line 22204: spectral library spelled wrong

psi-ms.obo
```diff
@@ -22196,6 +22198,68 @@ is_a: MS:1001456 ! analysis software
...
+
+[Term]
+id: MS:1003239
+name: library interpretation attribute set
+def: "A set of interpretation-related attributes that is shared by a subset of spectra within the same spectral library" [PSI:PI]
+is_a: MS:1003171 ! spectral library attribute
+
+[Term]
+id: MS:1003240
+name: peak intensity transforma
```
henryhlam, line 22233: should be "transform" not "transforma"

PR #107, opened 2022-03-11
Add proposed QC term spectra half-TIC

edeutsch:


mobiusklein (approved):

psi-ms.obo
```diff
@@ -22663,6 +22663,18 @@ relationship: has_value_type xsd\:float ! The allowed value-type for this CV ter
 relationship: has_value_concept NCIT:C25330 ! Duration
 relationship: has_units UO:0000010 ! second
 
+id: MS:4000068
```
mobiusklein, line 22667: Requires `[Term]` stanza opening syntax. I didn't see it before.

mobiusklein (approved):

psi-ms.obo
edeutsch: great, thanks for fixing!

mobiusklein:
@bittremieux if you can, please check that this conforms to the expected shape for mzQC

mwalzer (approved):
Thanks! This looks good.

bittremieux (approved):
Yes, looks good. Thanks!

PR #108, opened 2022-03-11
Update and add dissociation terms

edeutsch:


psi-ms.obo
```diff
@@ -22260,6 +22260,33 @@ def: "Accession number (e.g. in PeptideAtlas) of the peptide sequence" [PSI:PI]
...
+id: MS:1003245
+name: ultraviolet photodissociation
+def: "Multiphoton ionization where the reactant ion dissociates as a result of the absorption of multiple UV photons." [PSI:MS]
+synonym: "UVPD" EXACT []
+is_a: MS:1000435 ! photodissociation
+
+[Term]
+id: MS:1003246
+name: negative electron transfer dissociation
+def: "A process to fragment ions in a mass spectrometer by inducing fragmentation of anions (e.g. peptides or proteins) by transferring electrons to a radical-cation." [DOI:10.1016/j.jasms.2005.01.015] [PSI:MS]
```
mobiusklein, line 22273: The source of the definition should be a comma separated list enclosed in square braces:

```suggestion
def: "A process to fragment ions in a mass spectrometer by inducing fragmentation of anions (e.g. peptides or proteins) by transferring electrons to a radical-cation." [DOI:10.1016/j.jasms.2005.01.015, PSI:MS]
```

psi-ms.obo
```diff
@@ -22260,6 +22260,33 @@ def: "Accession number (e.g. in PeptideAtlas) of the peptide sequence" [PSI:PI]
...
+synonym: "NETD" EXACT []
+is_a: MS:1000044 ! dissociation method
+
+[Term]
+id: MS:1003247
+name: proton transfer reaction
+def: "A process transfer of one or more protons from a multiply charged cation (peptide or protein ion) to an anionic agent, that typically does not lead to fragmentation" [DOI: 10.1002/mas.20048] [PSI:MS]
```
mobiusklein, line 22280:
```suggestion
def: "A process transfer of one or more protons from a multiply charged cation (peptide or protein ion) to an anionic agent, that typically does not lead to fragmentation" [DOI: 10.1002/mas.20048, PSI:MS]
```

mobiusklein (changes requested):

psi-ms.obo
```diff
@@ -4242,7 +4242,7 @@ is_a: MS:1000462 ! ion optics
 [Term]
 id: MS:1000598
 name: electron transfer dissociation
-def: "A process to fragment ions in a mass spectrometer by inducing fragmentation of cations (e.g. peptides or proteins) by transferring electrons to them." [PSI:MS]
+def: "A process to fragment ions in a mass spectrometer by inducing fragmentation of cations (e.g. peptides or proteins) by transferring electrons from radical-anions." [DOI: 10.1073/pnas.0402700101, PMID: 15210983] [PSI:MS]
```
mobiusklein, line 4245:
```suggestion
def: "A process to fragment ions in a mass spectrometer by inducing fragmentation of cations (e.g. peptides or proteins) by transferring electrons from radical-anions." [DOI: 10.1073/pnas.0402700101, PMID: 15210983, PSI:MS]
```

mobiusklein:
Remaining ambiguity over whether the last two terms are "precursor activation" vs "dissociation method" which makes them separate.

chambm:
Dissociation method is:
relationship: part_of MS:1000456 ! precursor activation

Making the PTR terms direct is_a to precursor activation would make the relationships different.

But why don't we consider PTR to be an ionization method instead of an activation method?

edeutsch:
> But why don't we consider PTR to be an ionization method instead of an activation method?

I do not really know enough about PTR to say for sure, but what you suggest seems better. Appealing to @jshofstahl  and @pabinz for another opinion. Should
- proton transfer reaction
- negative proton transfer reaction
be children of MS:10000008 ionization type instead? See existing children here: (click on + to expand):
https://www.ebi.ac.uk/ols/ontologies/ms/terms?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FMS_1000008

pabinz:
To my understanding,

PTR-MS is originally designed to analyse volatile compounds. In this method a proton is transfered from a hydronium ion (H3O+) to a neutral analyte, leading to a protonated analyte. 

In proteomics, for instance in top down approach, ion/ion proton transfer (also named PTR) is used to reduce the number of protons attached to a multiply charge protein ion (and therefore simplifiy the ion cluster of a proteoform. The mechanism involves the presence of proton acceptors anions or neutral basic compounds that can capture protons from mutiply protonated protein ions. It is also used, in combination with ETD as fragmentation method, to target a reduction of the the complexity of the ion clusters in fragmentation spectra. In that respect, I would not consider it as activation method either.

PTR per se is not inducing fragmentation. 
In our CV, the use of the word activation is coupled with a dissociation/fragmentation event (CID, HCD, ETD, etc.).

Therefore PTR should be considered as an ionisation method and not an activation method.

And the definition should also consider a more generic description of a reaction that transfers one or more proton from a protonated molecule to a neutral or anionic proton acceptor species.

jshofstahl:
I was able to discuss the terms with John Syka yesterday and he did mention that PTR isn't the preferred name at this point due to the reason that Pierre-Alain mentioned.  We have switched to use Protein Transfer Charge Reduction (PTCR) for the technique in our documentation.  Some software still shows PTR inn the data but that is because they haven't switched to newer builds of our common components libraries.

edeutsch:
Based on the conversation, I have updated the PR to read like this now: Is that correct?

```
[Term]
id: MS:1003247
name: proton transfer reaction
def: "A process to transfer a proton from a hydronium ion (H3O+) to neutral analyte, leading to a protonated analyte, which typically does not lead to fragmentation" [DOI:10.1002/mas.20048, PSI:MS]
synonym: "PTR" EXACT []
is_a: MS:1000008 ! ionization type

[Term]
id: MS:1003248
name: proton transfer charge reduction
def: "A process to transfer of one or more protons from a multiply charged cation (peptide or protein ion) to a proton acceptor anion or neutral basic compound" [PSI:MS]
synonym: "PTCR" EXACT []
is_a: MS:1000008 ! ionization type
```

edeutsch:
Slight further edits. Comments?

```
[Term]
id: MS:1003247
name: proton transfer reaction
def: "Process to transfer a proton from a hydronium ion (H3O+) to neutral analyte, leading to a protonated analyte, which typically does not lead to fragmentation." [DOI:10.1002/mas.20048, PSI:MS]
synonym: "PTR" EXACT []
is_a: MS:1000008 ! ionization type

[Term]
id: MS:1003248
name: proton transfer charge reduction
def: "Process to transfer one or more protons from a multiply charged cation (peptide or protein ion) to a proton acceptor anion or neutral basic compound, thereby reducing the charge of the original analyte." [PSI:MS]
synonym: "PTCR" EXACT []
is_a: MS:1000008 ! ionization type
```

jshofstahl:
The most recent change from Eric looks OK to me.

mobiusklein (approved):

Issue #105, opened 2022-02-25
Automatic build failed - Should we remove NCIT?


edeutsch:
### Describe the question or discussion

It looks like the automatic generation of OWL failed with an out of memory error in the latest build attempt after today's merge.

I wonder if we should remove the import of NCIT. This is such a huge beast and I am doubtful that it is gaining us anything.

mobiusklein:
Will that break mzQC? Is there a restricted subset of NCIT that mzQC depends upon that has its own name?

I can try giving the JVM more memory, but the action worker only has 7 GB of RAM and I was at 4 GB previously.

bittremieux:
We use very few terms from NCIT: currently `Duration` and `Density`, in other mzQC terms that are not merged yet here also `Outlier` and `Cover`(age). These are value concepts that are not super essential, so if it causes issues the NCIT import could relatively safely be avoided I think.

I tried to find alternative terms in other ontologies we import (UO, STATO, PATO):

- [PATO:0001309](https://www.ebi.ac.uk/ols/ontologies/pato/terms?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FPATO_0001309): duration
- [STATO:0000333](https://www.ebi.ac.uk/ols/ontologies/stato/terms?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FSTATO_0000333): density -> kernel mixture distribution. Not ideal.
- [STATO:0000036](https://www.ebi.ac.uk/ols/ontologies/stato/terms?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FSTATO_0000036): outlier
- No alternative found for "coverage".

mwalzer:
Are there alternative converters that are less memory hungry? 
But yeah, NCIT is a big beast. But as a result of this size, there are a lot of terms for rich semantics in mzQC (don't know about others). 
In the meantime, we could try and get appropriate terms into STATO and PATO. As Wout mentioned, sometimes the definitions are not ideal.

mwalzer:
Pronto can nowadays dump OWL (according to this https://github.com/althonos/pronto/issues/149#issuecomment-1020471399). But I'm having difficulties loading any OBO after I updated...

mobiusklein:
[...]

PR #109, opened 2022-03-16
Add NCIT terms for mzQC

bittremieux:
As per #105, remove the full NCIT import and just re-declare the 4 NCIT terms that are used in mzQC.

I've used abridged terms from NCIT now that contain the relevant information, without some of the unnecessary `property_value` specifications. I'm not sure though whether the `is_a` specifications might be problematic now, because they refer to NCIT parent terms that are not included in our CV? It shouldn't be our goal to redefine the NCIT tree up to those terms, so if this is indeed problematic, what would be the solution? Have the terms consist of just `id`-`name`-`def`?

### DIFF
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -1,7 +1,7 @@
 format-version: 1.2
-data-version: 4.1.73
-date: 11:03:2022 02:22
-saved-by: Eric Deutsch
+data-version: 4.1.74
+date: 16:03:2022 10:37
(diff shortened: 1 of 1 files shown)


bittremieux:
The build was successful, so having those `is_a` terms doesn't seem to be a problem for that at least.

edeutsch (approved):

mobiusklein (approved):

PR #114, opened 2022-03-22
fixed QC terms syntax

mwalzer:
removed extra colons at relationship defs and corrected the dbxref to PSI:MS


mwalzer:
I think as we are sharing the accession prefix between regular PSI-MS and QC terms, it would be proper to also make the dbxref [PSI:MS]

mwalzer:
This PR is superseded by #115 (merging into another branch with more fixes like this)

Issue #116, opened 2022-03-24
identification counts


mwalzer:
### Describe the new term or terms you would like to add.

For an identification summary and statistics on groups of runs in mzQC, we'd like to introduce two terms in the spirit of https://github.com/HUPO-PSI/psi-ms-CV/blob/792582a7738540f5c3955b63b09dd3b0f2d91cb3/psi-ms.obo#L16464-L16470
for peptides and PSM. Closely following `MS:1002404` this would look like:
```
[Term]
id: MS:100xxxx
name: count of identified peptides
def: "The number of peptidoforms that have been identified and pass the threshold in the file." [PSI:PI]
xref: value-type:xsd\:int "The allowed value-type for this CV term."
is_a: MS:1002702 ! peptide sequence-level result list attribute
relationship: has_value_type xsd\:int ! The allowed value-type for this CV term

[Term]
id: MS:100xxxx
name: count of identified spectra
def: "The number of spectra that have been identified and pass the threshold in the file." [PSI:PI]
xref: value-type:xsd\:int "The allowed value-type for this CV term."
is_a: MS:1002700 ! PSM-level result list attribute
relationship: has_value_type xsd\:int ! The allowed value-type for this CV term
```

PR #117, opened 2022-03-24
adding peptide and spectrum identification count terms

mwalzer:
resolves https://github.com/HUPO-PSI/psi-ms-CV/issues/116

### DIFF
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -22319,8 +22319,8 @@ is_a: MS:1000008 ! ionization type
 
 [Term]
 id: MS:1003250
-name: count of identified peptides
-def: "The number of peptidoforms that have been identified and pass the threshold in the file." [PSI:PI]
+name: count of identified peptidoforms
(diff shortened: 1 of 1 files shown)


mobiusklein (approved):
Looks good to me.

bittremieux (approved):

edeutsch (approved):
This looks pretty good, I have a few nitpicky things:

1) Although the first time name is "count of identified peptides", the definition is "peptidoforms". This is thus ambiguous. Elsewhere in the CV we try to draw the distinction between "distinct peptide sequences" and "distinct peptidoforms" (which include mass mods). I suggest making the term name and definition match.

2) In both terms, it seems a little odd that the names are "...identified..." but the definitions are "...identified and pass threshold..." Maybe if would clearer if the definitions were:

"The number of xxxxxxxxxxx that pass the threshold to be considered identified with sufficient confidence."

edeutsch (approved):
nice, thanks!

PR #122, opened 2022-04-01
Add missing NCIT terms

### LINKED ISSUE #120: NCIT re-declaration (#109) works fine - almost
My apologies, I didn't have the opportunity to test the changes from #109 (using pronto). This was partly due to a [unit.obo](https://github.com/bio-ontology-research-group/unit-ontology/issues/44) issue that breaks loading of psi-ms.obo in pronto and needed a custom debug obo created. 
With that successfully circumvented, I realised that the re-declarations contain `is_a` statements that themselves contain NCIT terms _unknown_.
https://github.com/HUPO-PSI/psi-ms-CV/blob/d4d873eb049b7af007a4a2cdf02af97e216b07c9/psi-ms.obo#L89

Works fine without them, so can we ditch the `is_a` in these cases or would that make the concept re-declaration somehow invalid?

mobiusklein:
Fixes #120

### DIFF
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -1,5 +1,5 @@
 format-version: 1.2
-data-version: 4.1.80
+data-version: 4.1.81
 date: 01:04:2022 12:51
 saved-by: Joshua Klein
 auto-generated-by: OBO-Edit 2.3.1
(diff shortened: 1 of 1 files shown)


bittremieux (approved):

edeutsch (commented):
It seems that the "is_a"s of the terms in this PR were removed, presumably intentionally. In the official NCIT.obo, "Action" has this:
```
is_a: NCIT:C43431 ! Activity
```
which is omitted here.

When we discussed this I was envisioning adding Activity as well. It turns out Activity has no parent and is thus at the very top of the hierarchy.

It seem like there are 3 things we could do:
1) Include terms all the way up to the top of the NCIT. There aren't many since NCIT is shallow. e.g. Action has a parent Activity and that is that the top. This is what I envisioned.
2) Include immediate parents of the 4 original terms we wanted, and strip those parents of further parents (that is what this PR is)
3) Strip the 4 original terms we wanted of any parents so that we don't need any further terms. These 4 terms become top-level.

It seems to me that option 1 preserves the parts of NCIT that we want to include the best.

If we're going to strip out "is_a"s (like this PR does), I wonder why we strip the "is_a"s of these 4 terms rather than the 4 initial terms themselves?

I suppose I'm suggesting we do option 1, but I'm happy to be outvoted. I don't feel strongly.

edeutsch (approved):
Thanks, this is exactly what I envisioned and what I thought tidiest. This is good by me, thank you!

mobiusklein:
I'm fine either way. I can't test it against `pronto` because of a problem with `uo.obo`.

mwalzer (approved):
Yup, works (for me and in pronto)

PR #129, opened 2022-04-22
Add terms for mzSpecLib call 4/22/22

mobiusklein:


edeutsch (approved):
This looks pretty good to me. Two quibbles:
```
name: peak miscellaneous attribute
def: "An attribute of a peak in a mass spectrum other than its m/z, intensity, and annotation. " [PSI:PI]
```
I'm thinking that including "miscellaneous" is a bit icky. How about just "peak attribute". We could leave "peak intensity" where it currently is, and not rehomed here in case that were to cause a problem validating formats.

```
[Term]
id: MS:1003254
name: pre-decharged original charge state
```
I'm thinking that "original" is excessively redundant here. How about just "pre-decharged charge state"?

henryhlam:
> This looks pretty good to me. Two quibbles:
> 
> ```
> name: peak miscellaneous attribute
> def: "An attribute of a peak in a mass spectrum other than its m/z, intensity, and annotation. " [PSI:PI]
> ```
> 
> I'm thinking that including "miscellaneous" is a bit icky. How about just "peak attribute". We could leave "peak intensity" where it currently is, and not rehomed here in case that were to cause a problem validating formats.
> 
> ```
> [Term]
> id: MS:1003254
> name: pre-decharged original charge state
> ```
> 
> I'm thinking that "original" is excessively redundant here. How about just "pre-decharged charge state"?

henryhlam (changes requested):
Yes, let's make it "pre-decharged charge state" as Eric suggested.
For "peak miscellaneous attribute", I guess the plan was to use this to define the 4th column for the peak list. Not sure if "peak attribute" would be too broad for that purpose.

PR #132, opened 2022-04-27
DIA-NN added

ypriverol:


edeutsch (changes requested):
- The 4000000 series is for mzQC, the 1000000 series should be used for general terms like this
- Also, no changes to the OWL file are needed because the OWL is auto-genererated from the OBO
- Would you make the attribution instead of [PSI-MI] make it [PMID:31768060, https://github.com/vdemichev/DiaNN] instead? or another PMID or URL if more appropriate than my suggestion?

thanks!

mobiusklein:
The OWL changes are happening automatically because even on forks, commits to `master` run the automatic OWL file generation GitHub Action. They probably also try to tag and publish releases, but I don't know if it works.

edeutsch:
oh I see, if it's a PR from a local branch in this repo, that doesn't happen, but if it's a PR from a fork, then that will happen, got it. @ypriverol, scratch my comment about the OWL then. Not relevant. just my ignorance of GitHub.

ypriverol:
@edeutsch done!!

edeutsch (approved):
looks great to me, thanks!

mobiusklein (approved):
Looks good to me.

PR #137, opened 2022-05-21
More Types and Names from 5/20

mobiusklein:


edeutsch (approved):
all looks great to me, thanks!

Issue #130, opened 2022-04-25
Remove incorrect xrefs that are attempting to encode range constraints for properties-as-classes


cmungall:
none of the xrefs are correct

```
xref: SWO:http://edamontology.org/format_2187
xref: binary-data-type:MS\:1000519 "32-bit integer"
xref: binary-data-type:MS\:1000521 "32-bit float"
xref: binary-data-type:MS\:1000522 "64-bit integer"
xref: binary-data-type:MS\:1000523 "64-bit float"
xref: binary-data-type:MS\:1001479 "null-terminated ASCII string"
xref: value-type:xsd:string "The allowed value-type for this CV term."
xref: value-type:xsd\:anyURI "The allowed value-type for this CV term."
xref: value-type:xsd\:boolean "The allowed value-type for this CV term."
xref: value-type:xsd\:dateTime "The allowed value-type for this CV term."
xref: value-type:xsd\:decimal "The allowed value-type for this CV term."
xref: value-type:xsd\:double "The allowed value-type for this CV term."
xref: value-type:xsd\:float "The allowed value-type for this CV term."
xref: value-type:xsd\:int "The allowed value-type for this CV term."
xref: value-type:xsd\:integer "The allowed value-type for this CV term."
xref: value-type:xsd\:nonNegativeInteger "The allowed value-type for this CV term."
xref: value-type:xsd\:positiveInteger "Number of mantissa bits truncated."
xref: value-type:xsd\:positiveInteger "The allowed value-type for this CV term."
xref: value-type:xsd\:string "The allowed value-type for this CV term."
```

As I noted in

* #26.  

I think whatever you are trying to do here is better done in either OWL or a schema language like LinkML

With the exception of the SWO one (which should be encoded differently) these all seem to attempt to encode range constraints in obo format

```yaml
[Term]
id: MS:1000001
name: sample number
def: "A reference number relevant to the sample under study." [PSI:MS]
xref: value-type:xsd\:string "The allowed value-type for this CV term."
is_a: MS:1000548 ! sample attribute
relationship: has_value_type xsd\:string ! The allowed value-type for this CV term

[Term]
id: MS:1000002
name: sample name
def: "A reference string relevant to the sample under study." [PSI:MS]
xref: value-type:xsd\:string "The allowed value-type for this CV term."
is_a: MS:1000548 ! sample attribute
relationship: has_value_type xsd\:string ! The allowed value-type for this CV term

[Term]
id: MS:1000004
[...]

PR #141, opened 2022-07-01
Remove colon escape from relationship, not syntactically necessary

mobiusklein:
RE @chambm's comment on #130, this PR removes the colon escape from `xsd:` type declarations in relationships


chambm (approved):

edeutsch (approved):
it appears that my code related to has_value_type already strips out the whack if it exists. No problem if it disappears as far as I can tell. Good.

PR #143, opened 2022-07-13
Peak Annotation Terms Plus Extras

mobiusklein:
This PR adds the peak annotation terms discussed last Friday.

It also includes a few extras encountered during debugging:
1. We had not created CV terms for "other attribute name" and "other attribute value" for user-defined terms in mzSpecLib
2. The "scan number" term had no value-type definition
3. Several of the mass error related terms had "parts per notation" units, but not "parts per million" units

### DIFF
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -20354,33 +20354,33 @@ relationship: has_value_type xsd:string ! The allowed value-type for this CV ter
 
 [Term]
 id: MS:1003103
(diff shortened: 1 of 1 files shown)


edeutsch (commented):
Looks good to me, except for 'other attribute name' and 'other attribute value'. We should discuss before we go down that road.

edeutsch (approved):

PR #145, opened 2022-07-26
Update cross linking score regular expression and domain

### LINKED ISSUE #142: allow negative score in MS:1002665? ('regular expression for interaction scores derived from cross-linking')

Hi,
would it be possible to just change MS:1002665 -- 'regular expression for interaction scores derived from cross-linking' -- so that it allows a negative score?

These appear in our results sometimes and i was wondering if the validation issues it causes could be solved by just changing the reg ex to allow them,
best wishes,
Colin

mobiusklein:
Closes #142

This PR adds the requested changes to the regular expression for parsing cross linking scores in mzIdentML, and updates the expected domain for those terms bound to the regular expression to permit negative values.

@colin-combe if you could, please check that this matches your intent with the request?

### DIFF
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -1,6 +1,6 @@
 format-version: 1.2
-data-version: 4.1.94
-date: 07:01:2022 11:44
+data-version: 4.1.95
+date: 07:01:2022 11:53
 saved-by: Joshua Klein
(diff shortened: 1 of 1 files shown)


edeutsch (approved):
looks good, thanks! Maybe update the date at the top?

PR #146, opened 2022-08-12
Peak attribute terms for mzSpecLib

mobiusklein:
Adds the three new peak attribute terms describing how consistent a peak is, and makes the three scalar ion mobility terms peak attributes in addition to their other types.


edeutsch (commented):
instead of "frequency", which is usually time related, how about "fraction"?

name: observation fraction of peak
def: "The fraction of replicate spectra of the same analyte of which this peak is observed." [PSI:PI]

henryhlam:
> 

I like "frequency" better... I guess I think of it as a statistical term, not a physical one (e.g. frequency of getting a tail when we flip a coin many times) If we go for "fraction", maybe we need to be more clear like "fraction of replicates containing peak" as "observation fraction" does not seem to make sense to me.

edeutsch (approved):

henryhlam (approved):

PR #152, opened 2022-08-23
Add Bruker TSF format

chambm:
TSF is Bruker's replacement for the BAF format. It uses the same SDK as the TDF Format but has no ion mobility dimension. Each frame is a single scan, so we could make the nativeId `scan=`, but in the TSF schema it's called a frame.


edeutsch (approved):
Looks great, thanks, maybe a date and name change before the merge.

PR #155, opened 2022-09-07
New batch of QC metric terms

mwalzer:
* acquisition metrics
* some id-based metrics

### DIFF
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -22195,6 +22195,8 @@ comment: An unusual high count of signal jumps or falls can indicate ESI stabili
 is_a: MS:4000003 ! single value
 relationship: has_metric_category MS:4000009 ! ID free metric
 relationship: has_metric_category MS:4000021 ! MS1 metric 
+relationship: has_units UO:0000189 ! count unit
+relationship: has_value_type xsd:integer ! The allowed value-type for this CV term
 synonym: "IS-1A"  RELATED []
 
 [Term]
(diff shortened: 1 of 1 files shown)


edeutsch (commented):
After discussion at the call, we recommend improving the definitions of the first two terms to not just be repeats of the names. And also append the comments to the definitions.

psi-ms.obo
```diff
@@ -22315,6 +22321,102 @@ relationship: has_optional_column MS:4000087 ! mzQC plot label
...
+name: number of empty MS3 scans
+def: "Number of MS3 scans where the scans' peaks intensity sums to 0 (i.e. no peaks or only 0-intensity peaks)." [PSI:MS]
+is_a: MS:4000003 ! single value
+relationship: has_metric_category MS:4000009 ! ID free metric
+relationship: has_metric_category MS:4000012 ! single run based metric
+relationship: has_units UO:0000189 ! count unit
+relationship: has_value_type xsd:integer ! The allowed value-type for this CV term
+
+[Term] 
+id: MS:4000102
```
mwalzer, line 22401: We could of course put these outside the MS:4000000 scope to keep them with the other '`count of ..`' terms (MS:1001121,MS:1002404,MS:1002406,MS:1003059,MS:1003250,MS:1003251
)

mobiusklein (approved):
Requested changes were discussed on call and accepted.

PR #157, opened 2022-09-13
Add TSQ Altis Plus

chambm:


chambm:
One day I'll remember to bump the version in the same commit as the main change. :) Thanks for the quick review Eric.

chambm:
@mobiusklein Any idea why owl-tag-release is failing?

mobiusklein:
Yes, it looks like someone turned on branch protection on `master` (or it's now on by default), so the Action just pushing to master got blocked.

edeutsch:
thank you! I don't think I changed anything myself.

mobiusklein:
Options are to instead generate a new branch to merge for each OWL update, optionally programmatically approving and merging said branches, disabling branch protection, or giving a specific user the power to bypass branch protection and have the Action run as that user.

Option 1 is complicated. Option 2 is cumbersome. Option 3 seems to be counter to someone's wishes. So Option 4 it is. Since the Action in question runs with a token generated by me, all push actions it drives are done by my account, so I set the policy to not apply to me.

If we _really_ need to protect things from me too, we need to make a bot account to provide tokens.

edeutsch:
Many thanks @mobiusklein for addressing this. I think your option 4 seems like a good solution and I'm fine to continue with that.

PR #162, opened 2022-10-27
Spectrum match and spectral similarity terms

mobiusklein:


henryhlam (commented):
On spectral angle, perhaps we should instead define it as: "The angle subtended by two vectorized spectra in high-dimensional space. It is equal to the inverse cosine of the dot product."

edeutsch (approved):

PR #168, opened 2022-11-09
Add zstd compression term

### LINKED ISSUE #167: New Term Request: Zstd binary compression

Hi,  
I'd like to request a term for Zstd compression, analogous to zlib, but it can achieve greater performance.
http://facebook.github.io/zstd/

```
[Term]
id: MS:1000XXX
name: Zstd compression
def: "Zstd." [PSI:MS]
is_a: MS:1000572 ! binary data compression type
```

mobiusklein:
Closes #167 

@lazear I've added the term and a reference to the zstd website. Do you have any additional specifics you want to add like bit shuffling, dictionaries, or using it in tandem with one of the lossy compression schemes like those in Numpress?

### DIFF
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -19666,12 +19666,6 @@ is_a: MS:1000572 ! binary data compression type
 
 [Term]
 id: MS:1002995
-name: MS-Numpress short logged float compression followed by Zstd compression
-def: "Compression using MS-Numpress short logged float compression followed by zstd compression." [https://github.com/ms-numpress/ms-numpress]
-is_a: MS:1000572 ! binary data compression type
-
-[Term]
-id: MS:1002995
 name: Andromeda:PEP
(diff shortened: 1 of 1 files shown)


lazear:
I will defer to your expertise here - I think in light of Eric's valid concerns about portability, introducing more advanced things like dictionaries, shuffling, etc might just further compound the issue. Seems like numpress + zstd might make sense though?

chambm:
I suspect there are quite a few implementations that don't even support Numpress, so I suppose adding one more isn't a terrible idea. But seeing the small performance gap between intel's zlib and zstd, it doesn't seem like a big value add. I'm not aware that we have size problems with centroid data. For profile data, was there much of a difference in compression between zlib and zstd?

edeutsch:
decided that we should focus on mzMLb instead.

PR #172, opened 2022-12-09
distribution representation for various idfree metrics

mwalzer:
This PR introduces descriptive statistic metrics terms. Some (though not all of the included) items can already be represented with existing terms, this however an alternative/more condensed and flexible representation version of metrics.
Each distribution is represented by a term for the quartiles, for the standard deviation, the mean, and 'outliers' below or above 1.5*IQR. 
Adds terms for:
* peak densities
* S/N
* ion-collection times

### DIFF
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -22663,9 +22663,7 @@ comment: The distribution of peak densities in MS1 can provide insight into the 
 is_a: MS:4000004 ! n-tuple
 relationship: has_metric_category MS:4000009 ! ID free metric
 relationship: has_metric_category MS:4000021 ! MS1 metric
-relationship: has_value_concept STATO:0000167 ! first quartile
-relationship: has_value_concept STATO:0000574 ! center value
-relationship: has_value_concept STATO:0000170 ! third quartile
+relationship: has_value_concept STATO:0000291 ! quantile
 relationship: has_value_type xsd:float ! The allowed value-type for this CV term
 	
 [Term]
@@ -22777,9 +22775,7 @@ comment: The intensity distribution of the precursors informs about the dynamic 
 is_a: MS:4000004 ! n-tuple
 relationship: has_metric_category MS:4000009 ! ID free metric
 relationship: has_metric_category MS:4000022 ! MS2 metric
-relationship: has_value_concept STATO:0000167 ! first quartile
-relationship: has_value_concept STATO:0000574 ! center value
-relationship: has_value_concept STATO:0000170 ! third quartile
+relationship: has_value_concept STATO:0000291 ! quantile
 relationship: has_value_type xsd:float ! The allowed value-type for this CV term
 relationship: has_units MS:1000043 ! intensity unit
 
@@ -22835,9 +22831,7 @@ comment: The Signal-to-noise ratio in MS1 distribution can provide insight to th
 is_a: MS:4000004 ! n-tuple
 relationship: has_metric_category MS:4000009 ! ID free metric
 relationship: has_metric_category MS:4000021 ! MS1 metric
-relationship: has_value_concept STATO:0000167 ! first quartile
-relationship: has_value_concept STATO:0000574 ! center value
-relationship: has_value_concept STATO:0000170 ! third quartile
+relationship: has_value_concept STATO:0000291 ! quantile
 relationship: has_value_type xsd:float ! The allowed value-type for this CV term
 
 [Term]
@@ -22892,9 +22886,7 @@ comment: A high signal-to-noise ratio in MS2 distribution can explain a high rat
 is_a: MS:4000004 ! n-tuple
 relationship: has_metric_category MS:4000009 ! ID free metric
 relationship: has_metric_category MS:4000022 ! MS2 metric
-relationship: has_value_concept STATO:0000167 ! first quartile
-relationship: has_value_concept STATO:0000574 ! center value
-relationship: has_value_concept STATO:0000170 ! third quartile
+relationship: has_value_concept STATO:0000291 ! quantile
 relationship: has_value_type xsd:float ! The allowed value-type for this CV term
 	
 [Term]
@@ -22947,10 +22939,8 @@ name: MS1 ion collection time - quantiles
 def: "From the distribution of ion injection times (MS:1000927) for MS1, the quantiles. I.e. a value triplet represents the quartiles Q1, Q2, Q3" [PSI:MS]
 comment: Injection time distribution can be used to gauge the suitability of used instrument settings for the sample content used.
 is_a: MS:4000004 ! n-tuple
-relationship: has_value_concept STATO:0000167 ! first quartile
 relationship: has_metric_category MS:4000021 ! MS1 metric
-relationship: has_value_concept STATO:0000574 ! center value
-relationship: has_value_concept STATO:0000170 ! third quartile
+relationship: has_value_concept STATO:0000291 ! quantile
 relationship: has_units UO:0000028 ! millisecond
 relationship: has_value_type xsd:float ! The allowed value-type for this CV term
 
@@ -23010,9 +23000,7 @@ def: "From the distribution of ion injection times (MS:1000927) for MS2, the qua
 comment: Injection time distribution can be used to gauge the suitability of used instrument settings for the sample content used.
 is_a: MS:4000004 ! n-tuple
 relationship: has_metric_category MS:4000009 ! ID free metric
-relationship: has_value_concept STATO:0000167 ! first quartile
-relationship: has_value_concept STATO:0000574 ! center value
-relationship: has_value_concept STATO:0000170 ! third quartile
+relationship: has_value_concept STATO:0000291 ! quantile
 relationship: has_units UO:0000028 ! millisecond
 relationship: has_value_type xsd:float ! The allowed value-type for this CV term
 	
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -1,7 +1,7 @@
 format-version: 1.2
-data-version: 4.1.108
-date: 26:10:2022 11:38
-saved-by: Joshua Klein
+data-version: 4.1.110
+date: 09:12:2022 12:01
+saved-by: Mathias Walzer
 auto-generated-by: OBO-Edit 2.3.1
 import: http://purl.obolibrary.org/obo/pato.obo
 import: http://purl.obolibrary.org/obo/stato.owl
@@ -20682,7 +20682,7 @@ relationship: has_regexp MS:1002505 ! regular expression for modification locali
 [Term]
 id: MS:1003149
 name: PTMProphet normalized information content
-def: " PTMProphet-computed PSM-specific normalized (0.0 – 1.0) measure of information content across all modifications of a specific type." [DOI:10.1021/acs.jproteome.9b00205, PMID:31290668]
+def: " PTMProphet-computed PSM-specific normalized (0.0 - 1.0) measure of information content across all modifications of a specific type." [DOI:10.1021/acs.jproteome.9b00205, PMID:31290668]
 is_a: MS:1001968 ! PTM localization PSM-level statistic
 relationship: has_regexp MS:1002505 ! regular expression for modification localization scoring
 
@@ -21319,7 +21319,7 @@ is_a: MS:1003240 ! peak intensity transform
 [Term]
 id: MS:1003243
 name: adduct ion mass
-def: "The theoretical mass of the adduct ion (e.g. for a singly-charged protonated peptide ion, this value would be the neutral peptide molecule’s mass plus the mass of a proton)" [PSI:PI]
+def: "The theoretical mass of the adduct ion (e.g. for a singly-charged protonated peptide ion, this value would be the neutral peptide molecule's mass plus the mass of a proton)" [PSI:PI]
 is_a: MS:1003056 ! adduct ion property
 relationship: has_value_type xsd:float ! The allowed value-type for this CV term
 
@@ -22086,7 +22086,7 @@ relationship: has_units UO:0000010 ! second
 [Term]
 id: MS:4000054
 name: TIC quarters RT fraction
-def: "The interval when the respective quarter of the TIC accumulates divided by retention time duration." [PSI:QC]
+def: "The interval when the respective quarter of the TIC accumulates divided by retention time duration." [PSI:MS]
 synonym: "RT-TIC-Q1" RELATED [PMID:24494671]
 synonym: "RT-TIC-Q2" RELATED [PMID:24494671]
 synonym: "RT-TIC-Q3" RELATED [PMID:24494671]
@@ -22102,7 +22102,7 @@ relationship: has_units UO:0000191 ! fraction
 [Term]
 id: MS:4000055
 name: MS1 quarter RT fraction
-def: "The interval used for acquisition of the first, second, third, and fourth quarter of all MS1 events divided by retention time duration." [PSI:QC]
+def: "The interval used for acquisition of the first, second, third, and fourth quarter of all MS1 events divided by retention time duration." [PSI:MS]
 synonym: "RT-MS-Q1" RELATED [PMID:24494671]
 synonym: "RT-MS-Q2" RELATED [PMID:24494671]
 synonym: "RT-MS-Q3" RELATED [PMID:24494671]
@@ -22118,7 +22118,7 @@ relationship: has_units UO:0000191 ! fraction
 [Term]
 id: MS:4000056
 name: MS2 quarter RT fraction
-def: "The interval used for acquisition of the first, second, third, and fourth quarter of all MS2 events divided by retention time duration." [PSI:QC]
+def: "The interval used for acquisition of the first, second, third, and fourth quarter of all MS2 events divided by retention time duration." [PSI:MS]
 synonym: "RT-MSMS-Q1" RELATED [PMID:24494671]
 synonym: "RT-MSMS-Q2" RELATED [PMID:24494671]
 synonym: "RT-MSMS-Q3" RELATED [PMID:24494671]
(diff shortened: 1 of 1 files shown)


edeutsch (changes requested):
I'm thinking it would be good to follow the convention of the PSI-MS (and other QC terms) to not capitalize the first letter of the name unless it is a proper noun or an acronym. So for example, this:

Peak density distribution MS1 low outliers (<Q1-1.5*IQR)

would begin with a lower p

And also, I think it seems like the existing convention is to put the MS level first (most existing terms like MS:4000063 put the MS level first). So this might be better:

MS1 peak density distribution low outliers (<Q1-1.5*IQR)

And then I do wonder if it is prudent to put the exact threshold in the name. Do we really want the <Q1-1.5*IQR as part of the name? Maybe the concept of "outliers" is sufficient? and recommend <Q1-1.5*IQR in the definition?

I don't feel strongly about this last suggestion. If being this specific is most desirable, then fine.
But I do think that all the terms here should follow the first two conventions mentioned above.

psi-ms.obo
```diff
@@ -22655,6 +22655,386 @@ relationship: has_metric_category MS:4000012 ! single run based metric
 relationship: has_value_type xsd:int ! The allowed value-type for this CV term
 relationship: has_units UO:0000189 ! count unit
 
+[Term]
+id: MS:4000108
+name: Peak density distribution MS1 Q1, Q2, Q3
```
bittremieux, line 22660: **suggestion:** Should we rename this as "quantiles" and dynamically determine the number of splits from the values, as is the case for a few other terms already? In that case term `MS:4000109` would maybe be superfluous, because it would be the distribution consisting of only a single value?

psi-ms.obo
```diff
@@ -22655,6 +22655,386 @@ relationship: has_metric_category MS:4000012 ! single run based metric
...
+def: "From the distribution of peak densities in MS1, the quartiles Q1, Q2, Q3" [PSI:MS]
```
bittremieux, line 22661: **question:** Can you elaborate a bit about what the peak density values are? I don't immediately know how to interpret this description.

psi-ms.obo
```diff
@@ -22655,6 +22655,386 @@ relationship: has_metric_category MS:4000012 ! single run based metric
...
+def: "From the distribution of peak densities in MS1, the mean" [PSI:MS]
+comment: The distribution of peak densities in MS1 can provide insight into the presence of nuisance factors.
+is_a: MS:4000003 ! single value
+relationship: has_metric_category MS:4000009 ! ID free metric
+relationship: has_value_concept STATO:0000401 ! sample mean
+relationship: has_value_type xsd:float ! The allowed value-type for this CV term
+	
+[Term]
+id: MS:4000110
+name: Peak density distribution MS1 sigma
```
bittremieux, line 22682: **suggestion:** Rename "sigma" to "standard deviation" to be more clear?

psi-ms.obo
```diff
@@ -22655,6 +22655,386 @@ relationship: has_metric_category MS:4000012 ! single run based metric
...
+def: "From the distribution of peak densities in MS1, the sigma value" [PSI:MS]
+comment: The distribution of peak densities in MS1 can provide insight into the presence of nuisance factors.
+is_a: MS:4000003 ! single value
+relationship: has_metric_category MS:4000009 ! ID free metric
+relationship: has_value_concept STATO:0000237 ! standard deviation
+relationship: has_value_type xsd:float ! The allowed value-type for this CV term
+	
+[Term]
+id: MS:4000111
+name: Peak density distribution MS1 low outliers (<Q1-1.5*IQR)
```
bittremieux, line 22692: **suggestion:** I agree with @edeutsch that we probably don't need to specify how outliers are determined in the term name. Similarly, it might be sufficient to recommend using 1.5 IQR as the threshold for outliers, rather than require it. Ultimately, what is an outlier depends on the data distribution, and in some cases you might use different thresholds.

psi-ms.obo
```diff
@@ -22655,6 +22655,386 @@ relationship: has_metric_category MS:4000012 ! single run based metric
 relationship: has_value_type xsd:int ! The allowed value-type for this CV term
 relationship: has_units UO:0000189 ! count unit
 
+[Term]
+id: MS:4000108
+name: Peak density distribution MS1 Q1, Q2, Q3
+def: "From the distribution of peak densities in MS1, the quartiles Q1, Q2, Q3" [PSI:MS]
+comment: The distribution of peak densities in MS1 can provide insight into the presence of nuisance factors.
+is_a: MS:4000004 ! n-tuple
+relationship: has_metric_category MS:4000009 ! ID free metric
```
bittremieux, line 22664: **suggestion:** Add relationship to MS:4000021 (MS1 metric).

psi-ms.obo
```diff
@@ -22655,6 +22655,386 @@ relationship: has_metric_category MS:4000012 ! single run based metric
...
+relationship: has_value_concept STATO:0000036 ! outlier
+relationship: has_value_type xsd:float ! The allowed value-type for this CV term
+
+[Term]
+id: MS:4000113
+name: Peak density distribution MS2 Q1, Q2, Q3
+def: "From the distribution of peak densities in MS2, the quartiles Q1, Q2, Q3" [PSI:MS]
+comment: The distribution of peak densities in MS2 can provide insight into the instruments ion selction settings.
+is_a: MS:4000004 ! n-tuple
+relationship: has_metric_category MS:4000009 ! ID free metric
```
bittremieux, line 22716: **suggestion:** Add relationship to MS:4000022 (MS2 metric).

psi-ms.obo
```diff
@@ -22655,6 +22655,386 @@ relationship: has_metric_category MS:4000012 ! single run based metric
...
+relationship: has_value_concept STATO:0000036 ! outlier
+relationship: has_value_type xsd:float ! The allowed value-type for this CV term
+
+[Term]
+id: MS:4000113
+name: Peak density distribution MS2 Q1, Q2, Q3
+def: "From the distribution of peak densities in MS2, the quartiles Q1, Q2, Q3" [PSI:MS]
+comment: The distribution of peak densities in MS2 can provide insight into the instruments ion selction settings.
```
bittremieux, line 22714: **nitpick:** "selction" typo.

psi-ms.obo
```diff
@@ -22655,6 +22655,386 @@ relationship: has_metric_category MS:4000012 ! single run based metric
...
+relationship: has_value_concept STATO:0000036 ! outlier
+relationship: has_value_type xsd:float ! The allowed value-type for this CV term
+
+[Term]
+id: MS:4000118
+name: Precursor intensity distribution Q1, Q2, Q3
+def: "From the distribution of precursor intensities, the quartiles Q1, Q2, Q3" [PSI:MS]
+comment: The intensity distribution of the precursors informs about the dynamic range of the acquisition.
+is_a: MS:4000004 ! n-tuple
+relationship: has_metric_category MS:4000009 ! ID free metric
```
bittremieux, line 22768: **suggestion:** Add relationship to MS:4000022 (MS2 metric). I assume this is an MS2 metric, because even though the information comes from the MS1 level, without an MS2 scan, there is no notion of a precursor.

psi-ms.obo
```diff
@@ -22655,6 +22655,386 @@ relationship: has_metric_category MS:4000012 ! single run based metric
...
+relationship: has_value_type xsd:float ! The allowed value-type for this CV term
+relationship: has_units MS:1000043 ! intensity unit
+
+[Term]
+id: MS:4000123
+name: Signal-to-noise ratio in MS1 - Q1, Q2, Q3
+def: "From the distribution of signal-to-noise ratio in MS1, the quartiles Q1, Q2, Q3 value" [PSI:MS]
+comment: The Signal-to-noise ratio in MS1 distribution can provide insight to the presence of nuisance factors.
+is_a: MS:4000004 ! n-tuple
+relationship: has_metric_category MS:4000009 ! ID free metric
```
bittremieux, line 22825: **suggestion:** Add relationship to MS:4000021 (MS1 metric).

psi-ms.obo
```diff
@@ -22655,6 +22655,386 @@ relationship: has_metric_category MS:4000012 ! single run based metric
...
+relationship: has_value_concept STATO:0000036 ! outlier
+relationship: has_value_type xsd:float ! The allowed value-type for this CV term
+
+[Term]
+id: MS:4000128
+name: Signal-to-noise ratio in MS2 - Q1, Q2, Q3
+def: "From the distribution of signal-to-noise ratio in MS2, the quartiles Q1, Q2, Q3 value" [PSI:MS]
+comment: A high signal-to-noise ratio in MS2 distribution can explain a high rate of unidentified spectra. Comparisons will likely only make local sense where the same method of S/N calculation is applied (or overall the same software to calculate this metric, see mzQC 'analysisSoftware').
+is_a: MS:4000004 ! n-tuple
+relationship: has_metric_category MS:4000009 ! ID free metric
```
bittremieux, line 22877: **suggestion:** Add relationship to MS:4000022 (MS2 metric).

psi-ms.obo
```diff
@@ -22655,6 +22655,386 @@ relationship: has_metric_category MS:4000012 ! single run based metric
...
+relationship: has_value_concept STATO:0000036 ! outlier
+relationship: has_value_type xsd:float ! The allowed value-type for this CV term
+
+[Term]
+id: MS:4000133
+name: MS1 ion collection time Q1, Q2, Q3
+def: "From the distribution of ion injection times (MS:1000927) for MS1, the quartiles Q1, Q2, Q3" [PSI:MS]
+comment: Injection time distribution can be used to gauge the suitability of used instrument settings for the sample content used.
+is_a: MS:4000004 ! n-tuple
+relationship: has_value_concept STATO:0000167 ! first quartile
```
bittremieux, line 22929: **suggestion:** Add relationship to MS:4000021 (MS1 metric). Relationship to ID free metric is also missing here.

psi-ms.obo
```diff
@@ -22655,6 +22655,386 @@ relationship: has_metric_category MS:4000012 ! single run based metric
...
+relationship: has_units UO:0000028 ! millisecond
+relationship: has_value_type xsd:float ! The allowed value-type for this CV term
+
+[Term]
+id: MS:4000138
+name: MS2 ion collection time Q1, Q2, Q3
+def: "From the distribution of ion injection times (MS:1000927) for MS2, the quartiles Q1, Q2, Q3" [PSI:MS]
+comment: Injection time distribution can be used to gauge the suitability of used instrument settings for the sample content used.
+is_a: MS:4000004 ! n-tuple
+relationship: has_metric_category MS:4000009 ! ID free metric
```
bittremieux, line 22986: **suggestion:** Add relationship to MS:4000022 (MS2 metric).

bittremieux (changes requested):
I agree with all of Eric's suggestions.

For some of these terms I also don't immediately know what they mean or how they could be computed. Are there tools that calculate these values already? Maybe a sketch would be useful to clarify.

psi-ms.obo
mwalzer: :+1:

psi-ms.obo
mwalzer: :+1:

psi-ms.obo
mwalzer: I'd favour a short name and let `relationship: has_value_concept STATO:0000237 ! standard deviation` do the specification, but my heart's not in it.

psi-ms.obo
mwalzer: Let's mark down some time in the next telco to refine the definition

psi-ms.obo
mwalzer: I think in the box-whisker display 1.5IQR is pretty much the standard but I do *not* disagree with you, prob better to leave it out of the name. Leaving the threshold to the user might pose some serious issues downstream, let's also discuss in next telco

bittremieux: For e.g. submissions to Nature journals you now need to specify in all figure captions what the whiskers of box plots indicate, so it seems like the field is moving toward being more explicit rather than assuming that it's the default definition of 1.5 IQR. I found it a bit weird/annoying at first as well.

psi-ms.obo
mwalzer: Not impossible, but it would be good if we discuss the ways we handle that type of information for downstream purposes.

mwalzer: As discussed in telco, now quantiles, a visualisation guide to the metrics to follow (outside obo)

psi-ms.obo
mwalzer: i.e. you decide

psi-ms.obo
mwalzer: also deduplicated with quameter terms in existence already

edeutsch (changes requested):
The latest changes seem pretty good except for the dashes in the names. I suggest removing all the dashes. The term names are intended to be phrases as seen in a sentence. So I think that "precursor intensity distribution mean" is perfectly good without the dash. We can use it in a sentence: "For this file, the precursor intensity distribution mean was 1.72e7".

psi-ms.obo
```diff
@@ -22734,6 +22738,425 @@ relationship: has_metric_category MS:4000022 ! MS2 metric
...
+is_a: MS:4000003 ! single value
+relationship: has_metric_category MS:4000009 ! ID free metric
+relationship: has_metric_category MS:4000021 ! MS1 metric
+relationship: has_value_concept STATO:0000237 ! standard deviation
+relationship: has_value_type xsd:float ! The allowed value-type for this CV term
+	
+[Term]
+id: MS:4000124
+name: MS1 signal-to-noise ratio low outliers
+def: "From the distribution of signal-to-noise ratio in MS1, the list of outliers below Q1-1.5IQR" [PSI:MS]
```
bittremieux, line 22921: This should still be changed to "in-file defined threshold."

psi-ms.obo
```diff
@@ -22734,6 +22738,425 @@ relationship: has_metric_category MS:4000022 ! MS2 metric
...
+is_a: MS:4000004 ! n-tuple
+relationship: has_metric_category MS:4000009 ! ID free metric
+relationship: has_metric_category MS:4000021 ! MS1 metric
+relationship: has_value_concept STATO:0000036 ! outlier
+relationship: has_value_type xsd:float ! The allowed value-type for this CV term
+
+[Term]
+id: MS:4000125
+name: MS1 signal-to-noise ratio high outliers
+def: "From the distribution of signal-to-noise ratio in MS1, the list of outliers above Q3+1.5IQR" [PSI:MS]
```
bittremieux, line 22932: This should still be changed to "in-file defined threshold."

psi-ms.obo
```diff
@@ -22734,6 +22738,425 @@ relationship: has_metric_category MS:4000022 ! MS2 metric
...
+is_a: MS:4000003 ! single value
+relationship: has_metric_category MS:4000009 ! ID free metric
+relationship: has_metric_category MS:4000022 ! MS2 metric
+relationship: has_value_concept STATO:0000237 ! standard deviation
+relationship: has_value_type xsd:float ! The allowed value-type for this CV term
+	
+[Term]
+id: MS:4000129
+name: MS2 signal-to-noise ratio low outliers
+def: "From the distribution of signal-to-noise ratio in MS2, the list of outliers below Q1-1.5IQR" [PSI:MS]
```
bittremieux, line 22976: This should still be changed to "in-file defined threshold."

psi-ms.obo
```diff
@@ -22734,6 +22738,425 @@ relationship: has_metric_category MS:4000022 ! MS2 metric
...
+is_a: MS:4000004 ! n-tuple
+relationship: has_metric_category MS:4000009 ! ID free metric
+relationship: has_metric_category MS:4000022 ! MS2 metric
+relationship: has_value_concept STATO:0000036 ! outlier
+relationship: has_value_type xsd:float ! The allowed value-type for this CV term
+
+[Term]
+id: MS:4000130
+name: MS2 signal-to-noise ratio high outliers
+def: "From the distribution of signal-to-noise ratio in MS2, the list of outliers above Q3+1.5IQR" [PSI:MS]
```
bittremieux, line 22987: This should still be changed to "in-file defined threshold."

psi-ms.obo
```diff
@@ -22734,6 +22738,425 @@ relationship: has_metric_category MS:4000022 ! MS2 metric
...
+[Term]
+id: MS:4000142
+name: Tukey's fence
+def: "Defines outliers with Tuckey's fence as <(Q1-x*IQR) for low outliers and >(Q3+x*IQR) for high outliers, where x is defined by the term's value. The default is x=1.5" [PSI:MS]
+is_a: MS:4000141 ! outlier threshold criterion
+
+[Term]
+id: MS:4000143
+name: Tukey's fence high outliers
+def: "Defines high outliers with Tuckey's fence as >(Q3+x*IQR) for high outliers, where x is defined by the term's value. The default is x=1.5" [PSI:MS]
```
bittremieux, line 23127: Fix "Tuckey" to "Tukey" in a few different locations.

psi-ms.obo
```diff
@@ -22734,6 +22738,425 @@ relationship: has_metric_category MS:4000022 ! MS2 metric
...
+relationship: has_metric_category MS:4000009 ! ID free metric
+relationship: has_metric_category MS:4000021 ! MS1 metric
+relationship: has_value_concept STATO:0000291 ! quantile
+relationship: has_value_type xsd:float ! The allowed value-type for this CV term
+
+[Term]
+id: MS:4000122
+name: MS1 signal-to-noise ratio mean
+comment: The signal-to-noise ratio in MS1 distribution can provide insight into the presence of nuisance factors. Comparisons will likely only make local sense where the same method of S/N calculation is applied (or overall the same software to calculate this metric, see mzQC 'analysisSoftware').
+def: "From the distribution of signal-to-noise ratio in MS1, the mean" [PSI:MS]
```
bittremieux, line 22900: Move "def" above "comment".

bittremieux (changes requested):

psi-ms.obo
mwalzer: ta, missed that

psi-ms.obo
mwalzer: ta, missed that

psi-ms.obo
mwalzer: idk why it doesnt say outdated, fixed in be87214

ypriverol (approved):
Everything looks good.

bittremieux (approved):

edeutsch (approved):

PR #175, opened 2023-01-07
Add regex changes for modification localization and crosslinking FDR

### LINKED ISSUE #170: problems with MS:1002676 (Protein-pair level global FDR) and MS:1002677 (Residue pair-level global FDR)

Hi,
in short MS:1002676 (Protein-pair level global FDR) and MS:1002677 (Residue pair-level global FDR) have the wrong value type. 

They are strings described by a regular expression (see pg.22 of 1.2.0 mzId spec) but their value type is given as a double. This is also related to #150, you can see the problem in this image:
 
Screenshot from 2022-12-01 10-38-28

However, there are a couple of other things:

- our understanding of the mzIdentML spec is that CV terms giving scores can generally be used either to score a PSM or Peptide group, or to give the cut-off threshold for that score by including it in the Threshold element. Therefore, we think the regular expression for these terms (MS:1002665) should be changed to allow the possibility that it is just a number (for when MS:1002676 or MS:1002677 are placed in a Threshold element to give the cut-off score).
- we think the '+' or '-' sign in the score should be optional after the 'e' when using scientific notation (related #142) 

So, in additional to changing the value type of MS:1002676 and MS:1002677 to a string, we propose changing regex MS:1002665 from:
`([0-9]+)[.]([a|b]):([0-9]+|null):([\-+]?[0-9]+(?:[.][0-9]+)?(?:[Ee][+-][0-9]+)?):(true|false)`
to
`(([0-9]+)[.]([a|b]):([0-9]+|null):([-+]?[0-9]+(?:[.][0-9]+)?(?:[Ee][+-]?[0-9]+)?):(true|false))|([-+]?[0-9]+(?:[.][0-9]+)?(?:[Ee][+-]?[0-9]+)?)`
(have i got that right @lutzfischer )

feedback appreciated,
cheers,
Colin

### LINKED ISSUE #171: regex syntax for MS:1002505 (regular expression for modification localization scoring)

Hi,
@lutzfischer and I had a go at updating/fixing the regex syntax for MS:1002505 (regular expression for modification localization scoring).

We propose:
`([0-9]+):([01][.][0-9]+(?:[Ee][+\-]?[0-9]+)?):([0-9]+(?:[|][0-9]+)*):(true|false)`

we think this follows the syntax and use of capturing groups used by @mobiusklein in issue #142

The purpose of the regex is given in Section 5.2.8 of the mzIdentML 1.2.0 spec (pg. 17)
see also https://regex101.com/r/ZM1jTJ/1

mobiusklein:
Closes #171 
Closes #170 

@colin-combe did I miss anything from the set of issues?

### DIFF
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -1,6 +1,6 @@
 format-version: 1.2
-data-version: 4.1.109
-date: 22:12:2022 11:38
+data-version: 4.1.110
+date: 06:01:2023 11:38
 saved-by: Joshua Klein
 auto-generated-by: OBO-Edit 2.3.1
 import: http://purl.obolibrary.org/obo/pato.obo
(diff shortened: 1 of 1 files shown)


edeutsch (approved):

colin-combe:
> Closes #171 Closes #170
> 
> @colin-combe did I miss anything from the set of issues?

i think that's it

yes, i'd say its better not repeating the regular expression after, e.g.,  `relationship: has_regexp MS:1002665` but instead calling it by its name

thanks Josh,
C

PR #182, opened 2023-03-10
Add fixes from 2/28/23 crosslinking call

mobiusklein:
@colin-combe

### DIFF
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -1,6 +1,6 @@
 format-version: 1.2
-data-version: 4.1.114
-date: 17:02:2023 11:30
+data-version: 4.1.115
+date: 09:03:2023 11:30
 saved-by: Joshua Klein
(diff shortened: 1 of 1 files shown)


edeutsch (approved):
good, but maybe refine:
posterior error probability from identification based on multiple spectra
to
posterior error probability for an identification based on multiple spectra

edeutsch (commented):
Or better to use reverse decreasing scope notation instead of a preposition:
multi-spectrum identification posterior error probability

colin-combe:
i think one of @edeutsch suggestions is better/more correct - 
"posterior error probability for an identification based on multiple spectra" 
is better than
"posterior error probability from identification based on multiple spectra"

I think "multi-spectrum identification posterior error probability" is possibly less clear?
I was interested in "reverse decreasing scope notation", I googled it to see how it was defined but found nothing

PR #183, opened 2023-03-10
New high-level experiment types

edeutsch:


javizca (approved):
I would remove the bit small fragments called peptides, to just say peptides (in the term M:1003355), but this is a very small detail and happy to leave it unchanged.

edeutsch:
> I would remove the bit small fragments called peptides, to just say peptides (in the term M:1003355), but this is a very small detail and happy to leave it unchanged.

yes, good point, fixed.

mobiusklein (approved):

PR #186, opened 2023-03-28
dehyphenate looplink

colin-combe:
dehyphenate "looplink", as per several comments that said it was inconsistent with the dehyphenated "crosslink"


edeutsch (approved):

edeutsch (commented):
Need to update the version tag and date

mobiusklein:
Addressed in #189

PR #190, opened 2023-04-17
Add specific mass array types for top-down MS encoding

mobiusklein:
Discussion last year that @dtabb73 initiated suggested that the neutral mass array type is insufficient to describe the range of data types for top-down MS. Sometimes we may want to talk about monoisotopic mass, while in other scenarios we may want to describe the most abundant isotope's mass and there was no way to express this.

Alternatives:
Different units?
```xml
<binaryDataArray>
     <cvParam accession="MS:1000514" cvRef="PSI-MS" name="m/z array" unitAccession="MS:1000???" unitCvRef="PSI-MS" unitName="monoisotopic m/z" value=""/>
     ...
</binaryDataArray>  
```
Extra CV term?
```xml
<binaryDataArray>
     <cvParam accession="MS:1000???" cvRef="PSI-MS" name="most abundant m/z summarization" />
     <cvParam accession="MS:1000514" cvRef="PSI-MS" name="m/z array" unitAccession="MS:1000040" unitCvRef="PSI-MS" unitName="m/z" value=""/>
     ...
</binaryDataArray>
```

There are more complex array types to be discussed later.

### DIFF
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -1,5 +1,5 @@
 format-version: 1.2
-data-version: 4.1.118
+data-version: 4.1.119
 date: 31:03:2023 11:30
 saved-by: Joshua Klein
 auto-generated-by: OBO-Edit 2.3.1
@@ -22187,6 +22187,33 @@ name: tailor score
 def: "A calibrated version of the XCorr score, computed by dividing the XCorr by the 99th percentile of the distribution of all scores for a particular spectrum." [PMID:32175744]
 is_a: MS:1001143 ! PSM-level search engine specific statistic
 relationship: has_value_type xsd:double ! The allowed value-type for this CV term
+
+[Term]
+id: MS:1003367
+name: monoisotopic mass array
+def: "A data array of deisotoped neutral mass values corresponding to the estimated monoisotopic masses of the measured ions." [PSI:MS]
+xref: binary-data-type:MS\:1000521 "32-bit float"
+xref: binary-data-type:MS\:1000523 "64-bit float"
+is_a: MS:1003143 ! mass array
+relationship: has_units UO:0000221 ! dalton
+
+[Term]
+id: MS:1003368
+name: most abundant mass array
+def: "A data array of deisotoped neutral mass values corresponding to the mass of the most abundant isotopic peak of the measured ions." [PSI:MS]
+xref: binary-data-type:MS\:1000521 "32-bit float"
+xref: binary-data-type:MS\:1000523 "64-bit float"
+is_a: MS:1003143 ! mass array
+relationship: has_units UO:0000221 ! dalton
+
+[Term]
+id: MS:1003369
+name: average mass array
+def: "A data array of deisotoped neutral mass values corresponding to the mass of the most average mass of the measured ions." [PSI:MS]
+xref: binary-data-type:MS\:1000521 "32-bit float"
+xref: binary-data-type:MS\:1000523 "64-bit float"
+is_a: MS:1003143 ! mass array
+relationship: has_units UO:0000221 ! dalton
 
 [Term]
 id: MS:4000000
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -424,7 +424,7 @@ relationship: has_value_type xsd:string ! The allowed value-type for this CV ter
 [Term]
 id: MS:1000033
 name: deisotoping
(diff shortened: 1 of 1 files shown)


chambm:
We have the non-standard data array which can be used for assigning a mass type to indivdual peaks. What other different mass "data types" are there for top-down MS?

mobiusklein:
@chambm those other array types aren't mass arrays, but different ways of encoding information that is otherwise lost when you collapse an isotopic envelope down to a single mass peak.

The desire here is to express the deconvolved mass type for an entire spectrum formally, which'd be an improvement over the ambiguity you get in several top-down tools' text file outputs. There isn't a clear use case to express a single ion in average mass and the rest in monoisotopic mass (that came up in conversation).

chambm:
Seems like dataProcessing then:

id: MS:1000033
name: deisotoping
def: "The removal of isotope peaks to represent the fragment ion as one data point and is commonly done to reduce complexity. It is done in conjunction with the charge state deconvolution." [PSI:MS]
is_a: MS:1000543 ! data processing action

[Term]
id: MS:1000034
name: charge deconvolution
def: "The determination of the mass of an ion based on the mass spectral peaks that represent multiple-charge ions." [PSI:MS]
is_a: MS:1000543 ! data processing action

mobiusklein:
So instead of specifying this as an array type, express it as a specific type of data processing action that changes the semantics of the `mass array` and reference the appropriate `dataProcessing[id]` in the `binaryDataArray[dataProcessingRef]`. 

That makes more sense, and would let tools that don't know about those new array types run (and produce possibly nonsensical results) instead of crashing and burning because they cannot find the "expected" array type.

chambm:
Is "decharging" a word used in the top-down domain? The original wording was more plain, but if decharging is in use, it's fine.

mobiusklein:
I don't know, it was used elsewhere in the CV (but related spectral libraries). I'm happy to go back to the old name and add "decharging" as a synonym.

KyowonJeong:
> I don't know, it was used elsewhere in the CV (but related spectral libraries). I'm happy to go back to the old name and add "decharging" as a synonym.

Deconvolution is usually decharging + deisotoping.  where decharging is to assign the charges to isotope peaks and deisotoping is to find monoisotope masses. For some tools like UniDec that processes isotopically unresolved peaks, only decharing is done and monoiso masses are not reported. But for most deconvolution tools, they somehow try to do deisotoping as well.

KyowonJeong:
> 

To add a bit of details, a single mass in topdown often has many peaks that represent the mass with different isotope indices and charges. For example, a monoiso mass of 120 would have its isotopes 121 122 123Das. On top of that this mass was ionized with charge ranges from 2-6. Then the mass 120 would have its associated peaks  with m/zs (when ignoring charge carrier mass)

z=2:  120/2, 121/2, 1213
z=3: 120/3,121/3, 122/3
z=4: 120/4,121/4,122/4
z=5: 120/5.121/5,122/5
total 12 peaks in the raw spectrum.    Many deconvolution tools try to find this full association but they tend to be inconsistent, even if they reported the same mass 120. So for best reproducibility, it matters to store this mass to peaks association in any format. And I believe standard format for deconvolution output should be introduced very soonish since now the TDP society is accelerating to come up with standardized protocols and analysis. CTDP (consortium for top down) is launching a new initiative on standardization.

mobiusklein:
@KyowonJeong my understanding is that `charge deconvolution` is assigning charge states to isotopic envelopes, and that `decharging` goes one step further to transform isotopic envelopes (or collapsed single peaks if deisotoping was done) from `z=x` to `z=1` and to merge peaks that are so close as to be indistinguishable, effectively discarding information about the relative abundances of different charge states.

Is this distinction correct?

KyowonJeong:
> 
Your distinction is very clear and I would like to advocate the definitions. But I have not seen many literature mentioning "charge deconvolution" in top down area. Mostly we say decharging to refer to assigning charges to peaks and simply moving the peaks from m/z to mass dimension without merging indistinguishable peaks into a single one (as in your definition of charge deconvolution). Some deconvolution methods merge such indistinguishable peaks into a single peak in the mass dimension but some do not - they treat isotope envelopes in different charges separately. So including this merging step into the term decharging may be not preferred in some contexts.

edeutsch (approved):
thanks for the PR and discussion. This seems reasonable to me.

PR #191, opened 2023-04-22
Add SelexION compensation voltage

### LINKED ISSUE #178: Add "SelexION differential mobility separation device"

Hi, we need a term `SelexION differential mobility separation device` .

### LINKED ISSUE #179: Add "differential mobility"

We need a term `differential mobility`.

mobiusklein:
Closes #179
Closes #178

### DIFF
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -10184,6 +10184,7 @@ name: FAIMS compensation voltage
 def: "The DC potential applied to the asymmetric waveform in FAIMS that compensates for the difference between high and low field mobility of an ion." [PSI:MS]
 synonym: "FAIMS CV" EXACT []
 is_a: MS:1002892 ! ion mobility attribute
+is_a: MS:1000455 ! ion selection attribute
 is_a: MS:1003254 ! peak attribute
 relationship: has_units UO:0000218 ! volt
 relationship: has_value_type xsd:double ! The allowed value-type for this CV term
@@ -22217,6 +22218,7 @@ id: MS:1003371
 name: SelexION compensation voltage
 def: "The voltage applied in the SelexION device to allow certain ions to transmit through to the mass spectrometer." [PSI:MS]
 is_a: MS:1002892 ! ion mobility attribute
+is_a: MS:1000455 ! ion selection attribute
 is_a: MS:1003254 ! peak attribute
 relationship: has_units UO:0000218 ! volt
 relationship: has_value_type xsd:float ! The allowed value-type for this CV term
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -1,5 +1,5 @@
 format-version: 1.2
-data-version: 4.1.119
+data-version: 4.1.120
 date: 31:03:2023 11:30
 saved-by: Joshua Klein
 auto-generated-by: OBO-Edit 2.3.1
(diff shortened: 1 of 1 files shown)


edeutsch (approved):
I approve, but then I wrote it. @chambm ? seem good?

chambm:
The definition seems very close to FAIMS CV:
```
[Term]
id: MS:1001581
name: FAIMS compensation voltage
def: "The DC potential applied to the asymmetric waveform in FAIMS that compensates for the difference between high and low field mobility of an ion." [PSI:MS]
synonym: "FAIMS CV" EXACT []
is_a: MS:1002892 ! ion mobility attribute
is_a: MS:1003254 ! peak attribute
relationship: has_units UO:0000218 ! volt
relationship: has_value_type xsd:double ! The allowed value-type for this CV term
```
but in simpler terms. So close that I wonder if they should be synonyms. And if they're not synonyms, could the definition be improved to clarify the difference? And in the latter case, should they both be under a parent term "differential mobility"?

mobiusklein:
From reading the paper on the principles be exploited here, it "looks like" they are the same, but I don't believe that they are on the same unit/scale of separation, just that the settings for both devices happens to be in units of compensation voltage. I wouldn't treat them as synonyms.

I agree with making a common parent. I just don't have the first principles knowledge to know I'd describe it well but I can take a shot at it.

edeutsch:
From my reading the SelexION device was described as being a bit different, a simpler design. One option is that we just throw SelexION in with FAIMS. I'm thinking that the companies would like to have them separate, but I don't really know. It seems harmless to keep them separate. Putting them together under a single parent term "differential ion mobility device setting" seems reasonable, although I wonder if it necessary.

chambm:
I agree companies might prefer to see them separate. So you think the current definitions reflect the simpler design? That's what I'm not sure about. "DC potential" is this context is synonymous with "voltage". Does SelexION not deal with the "high and low field mobility of an ion"? 

Substituting synonyms and removing the device name, I see the definitions as:
"The DC potential applied to allow certain ions to transmit through to the mass spectrometer." 
"The DC potential applied to the asymmetric waveform that compensates for the difference between high and low field mobility of an ion."

edeutsch:
> So you think the current definitions reflect the simpler design? That's what I'm not sure about.

I don't know either. I simplified the definition in a way that I thought it was probably not wrong. I opted for "voltage" instead of "DC potential" since I wasn't sure if it was purely DC. AC can have a voltage too. I wasn't sure there is an asymmetric waveform either, so I removed that. I went with a functional definition that I thought was not likely wrong rather than a technical one that might be wrong.

In any case, I'm happy to go with whatever you think is best, thanks!

mobiusklein:
After considering it further, I do not think we need an extra parent term. I don't see the functional value as there is no use-case where knowing "this is an ion mobility attribute but not `ion mobility drift time` or `inverse reduced ion mobility`, one with a unit of volts, neither time nor volts/distance * time^2" because despite them both being voltage-dependent, they are still going to represent different units of separation.

While I was fiddling with it, I realized that FAIMS and this new term lacked `is_a: MS:1000455 ! ion selection attribute` while `ion mobility drift time` and `inverse reduced ion mobility` did have it. I don't see a reason why not to include that property, albeit that because of the way we read FAIMS data right now presumes that MS2 spectra acquired from an MS1 spectrum for a given CV inherit the CV of their parent scan. Were we to use a 4D representation like we do for higher granularity IM-MS data, we'd want the ability to say which CV segment a precursor came from.

edeutsch (approved):

PR #192, opened 2023-05-05
Add extension document versioning

mobiusklein:
Adds a new set of terms for expressing that a particular file conforming to a specification requires a specific extension document also be supported in order to be read properly.

This is in support of the extension documents under revision for mzIdentML.

### DIFF
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -22222,7 +22222,7 @@ relationship: part_of MS:0000000 ! Proteomics Standards Initiative Mass Spectrom
 id: MS:1003373
 name: mzIdentML extension version
 def: "The versioning of an mzIdentML extension document." [PSI:PI]
-relationship: part_of MS:0000000 ! Proteomics Standards Initiative Mass Spectrometry Vocabularies
+is_a: MS:1003372 ! mzIdentML extension version
 relationship: part_of MS:1002073 ! mzIdentML format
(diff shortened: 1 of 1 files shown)


psi-ms.obo
```diff
@@ -22212,6 +22212,20 @@ name: reduction to summed singly charged peak list
...
+id: MS:1003372
+name: specification document extension version
+def: "The versioning of a an extension to a specification document that the current file requires to be read correctly. The version should encode the name of the extension, and some ordinal expression of its revision, preferably in semantic versioning notation. Signals that readers that do not know this extension should return an appropriately informative error if they do not think they can or should try to interpret the file." [PSI:MS]
+relationship: part_of MS:0000000 ! Proteomics Standards Initiative Mass Spectrometry Vocabularies
+
+[Term]
+id: MS:1003373
+name: mzIdentML extension version
+def: "The versioning of an mzIdentML extension document." [PSI:PI]
+is_a: MS:1003372 ! mzIdentML extension version
```
edeutsch, line 22225: I think this is supposed to be:
```
is_a: MS:1003372 ! specification document extension version
```

?

edeutsch (changes requested):
text does not quite match the listed identifier. otherwise okay.

edeutsch (approved):

PR #201, opened 2023-06-19
Crosslinking extension document version

### LINKED ISSUE #196: new term for version of crosslinking extension document

Hi Joshua (@mobiusklein),
assuming I've correctly understood how the new MS:1003372 and MS:1003373 terms are meant to work, 
and bearing in mind the text you wrote in the draft 1.3.0 spec: "A document signals which extensions it requires by including a cvParam that is a child of MS:1003373",
then I think the new CV term for the version of the crosslinking extension document is yet to be created?

I think we need a new CV term, child of MS:1003373, with name "mzIdentML crosslinking extension document version", or similar?

We could consider making the value of MS:1003373 a regex that describes semantic versioning rather than a string. Or do you think that's overkill?

Thanks for your help,
Colin

mobiusklein:
Closes #196

### DIFF
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -1,5 +1,5 @@
 format-version: 1.2
-data-version: 4.1.124
+data-version: 4.1.125
 date: 04:05:2023 11:30
 saved-by: Joshua Klein
 auto-generated-by: OBO-Edit 2.3.1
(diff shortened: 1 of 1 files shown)


javizca:
all fine

edeutsch (approved):
Looks good to me. I think semantic version strings can be followed by a dash and string, so I think:
v1.2.3-dev
v2.3.4-beta2
are legal semvers. I don't know if we want to support that. Fine with me to not include it until necessary

PR #207, opened 2023-07-14
Add new crosslinking terms from most recent revision

mobiusklein:
@colin-combe


psi-ms.obo
```diff
@@ -22350,6 +22350,33 @@ name: mzTab-M
 def: "Expanded tabular result format for metabolomics experiments reporting quantitative summary data, MS features and identification evidence." [PMID:30688441, PMID:31525911, https://www.psidev.info/mztab]
 is_a: MS:1000914 ! tab delimited text format
 
+[Term]
+id: MS:1003390
+name: crosslinker cleavage
```
edeutsch, line 22355: I'm thinking that "crosslinker cleavage characteristics" is a better name for the intent of this concept?

psi-ms.obo
```diff
@@ -22350,6 +22350,33 @@ name: mzTab-M
...
+def: "Signifies that the crosslinker is cleavable and on cleavage can leave a given stub. The pattern specifies up to three slots <name>:<mass>:<pairs with>." [PSI:PI]
+is_a: MS:1001471 ! peptide modification details
+relationship: has_regexp MS:1003391 ! crosslinker cleavage regular expression
```
edeutsch, line 22358: Do we also need a line of relationship: has_value_type xsd:string here? Or do readers understand that has_regexp implies that? I'm thinking we need to be explicit about its value_type?

edeutsch (commented):
I suggest a few changes as shown below.

psi-ms.obo
colin-combe: yes, that is perhaps better, 
@lutzfischer / @lars-kolbowski - any comments?

colin-combe:
thanks Joshua

psi-ms.obo
lutzfischer: I am not sure in how far that is more accurate. If you ask me to characterise a crosslinker cleavage behaviour I would think, more then just the observable stubs play in; like e.g. under what conditions it cleaves. 
Now you could claim the same for describing "crosslink cleavage". So guess what I want to say is - I don't see the benefit of one vs the other. 
In summary I am fine with either name.

psi-ms.obo
lutzfischer: would it make sense to add
relationship: has_value_type xsd:string
to id: MS:1002479?
```
[Term]
id: MS:1002479
name: regular expression
def: "Regular expression." [PSI:PI]
relationship: part_of MS:0000000 ! Proteomics Standards Initiative Mass Spectrometry Vocabularies
relationship: has_value_type xsd:string
```
then every regex would have that

psi-ms.obo
```diff
@@ -22350,6 +22350,33 @@ name: mzTab-M
 def: "Expanded tabular result format for metabolomics experiments reporting quantitative summary data, MS features and identification evidence." [PMID:30688441, PMID:31525911, https://www.psidev.info/mztab]
 is_a: MS:1000914 ! tab delimited text format
 
+[Term]
+id: MS:1003390
+name: crosslinker cleavage
+def: "Signifies that the crosslinker is cleavable and on cleavage can leave a given stub. The pattern specifies up to three slots <name>:<mass>:<pairs with>." [PSI:PI]
```
lars-kolbowski, line 22356: Aren't all three slots mandatory?
```suggestion
def: "Signifies that the crosslinker is cleavable and on cleavage can leave a given stub. The pattern specifies three slots <name>:<mass>:<pairs with>." [PSI:PI]
```

psi-ms.obo
```diff
@@ -22350,6 +22350,33 @@ name: mzTab-M
...
+is_a: MS:1001471 ! peptide modification details
+relationship: has_regexp MS:1003391 ! crosslinker cleavage regular expression
+
+[Term]
+id: MS:1003391
+name: crosslinker cleavage regular expression
+def: "[A-Za-z]:[+-]?[0-9]+(\.[0-9]+)?([eE][+-]?[0-9]+(\.[0-9]+)?)?:[A-Za-z]+" [PSI:PI]
```
lars-kolbowski, line 22363: @colin-combe In your email from 7/12/2023 you stated that you updated the regexp to:
`^(?<NAME>[A-Za-z]):(?<SCORE>[+-]?[0-9]+(\.[0-9]+)?([eE][+-]?[0-9]+(\.[0-9]+)?)?):(?<PAIRS_WITH>[A-Za-z]+)$`
Did this change again or is this outdated here?

psi-ms.obo
lars-kolbowski: I'm also fine with either

psi-ms.obo
lutzfischer: I guess that should rather be `^(?<NAME>[A-Za-z]):(?<MASS>[+-]?[0-9]+(\.[0-9]+)?([eE][+-]?[0-9]+(\.[0-9]+)?)?):(?<PAIRS_WITH>[A-Za-z]+)$`

lars-kolbowski: :+1:  you're right should be `MASS` of course

colin-combe: yeah, stupid mistake from me

psi-ms.obo
colin-combe: if @edeutsch is the only one with a preference and others don't mind then lets go with @edeutsch preference.

>  I don't see the benefit of one vs the other.

I feel like "crosslinker cleavage" sounds more like something that happened, and "crosslinker cleavage characteristics" sounds more like a description of what might happen, so that's why i thought the latter might be better?

edeutsch (approved):

colin-combe (approved):
thanks

PR #218, opened 2023-10-11
Minor changes for MS2 and MS3

ypriverol:


ypriverol:
This PR is related with an SDRF issue where users want to define the analyzer and dissociation method specifically for MS2 and MS3.  https://github.com/bigbio/proteomics-sample-metadata/issues/679 . 

It would like to **know suggestions** what happen to the terms under `dissociation method` as child terms if they need to be added also as child terms of the new terms MS2 and MS3. Similarly, is the case for mass analyzer.

bittremieux (changes requested):
- I don't think these terms need to be in the QC namespace (i.e. `MS:4000XXX`).
- Do we then also need an explicit term for MS1 mass analyzer and dissociation method, or do the parent terms cover those?
- I don't fully understand your question about child terms. Can you please elaborate?

chambm:
I'm afraid it doesn't make sense to connect the analyzer to the ms level at all. It's very common for a run to use two dissociation methods in alternating MS2 scans. Likewise it's common for MS2 and MS3 to use the same dissociation method (that's the most common case). Dissociation method is thus a scan by scan property, not a run-level one.

ypriverol:
> I'm afraid it doesn't make sense to connect the dissociation method to the ms level at all. It's very common for a run to use two dissociation methods in alternating MS2 scans. Likewise it's common for MS2 and MS3 to use the same dissociation method (that's the most common case). Dissociation method is thus a scan by scan property, not a run-level one.

Hi @chambm I'm not really an expert but looks like it is possible that MS2 spectra are captured in one dissociation method and MS3 in another one. In those cases in the SDRF we can not use only the generic term dissociation method because we can't know if it was for the MS2 or MS3 spectra. I agreed we can make it generic and add two disassociation methods and the readers of the SDRF will need to figure out the relation between the MS level and the method used. My intention was to make clear that relation. But, we can discuss that on Friday's call and see the opinions of others. Other proposals about how to capture that are also welcome. 

@bittremieux I didn't know QC terms as jump IDs, I will change that in my PR. For MS1 I guess we can use the generic term, as @chambm mentioned I think the majority of the SDRF annotations will be using that generic term for all spectra.

About the children terms: When you have an ontology, for example within the dissociation method term you have children terms of all the possible methods. This enables tools, for example, to look up the parent terms for HCD (https://www.ebi.ac.uk/ols/ontologies/ms/terms?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FMS_1002481&lang=en&viewMode=All&siblings=false) and see that is a dissociation method. In principle, we should do the same for the MS2 and MS3 methods.

edeutsch:
I am thinking that the analyzer and dissociation methods are encoded in the raw or mzML files and the patterns can be simple or very complex within one MS run. As such, I don't think this information should be encoded in the SDRF file. I think we should reserve the SDRF file to encode information that cannot be encoded in the RAW/mzML files, like sample information.

ypriverol:
I will close the PR thanks for the feedback @chambm @edeutsch @bittremieux. Just a couple of thoughts: 

SDRF some data fields are optional, but they are perfectly possible to be annotated. The idea is that some users without the need to go into thousands of non-converted files can be able to filter experiments and samples by the data fields, for example: Give me all samples/data files HCD or with this particular PTM or for this particular instrument. This information can help a lot with the querying and descriptions at the dataset/sample level.

bittremieux:
I would argue that the latter use case should use mzQC instead, where we have a mechanism to store method at a scan, run, and experiment level. Ultimately, each dataset could have relevant metadata in mzQC file(s) to easily find relevant public data.

chambm:
> I will close the PR thanks for the feedback @chambm @edeutsch @bittremieux. Just a couple of thoughts:
> 
> SDRF some data fields are optional, but they are perfectly possible to be annotated. The idea is that some users without the need to go into thousands of non-converted files can be able to filter experiments and samples by the data fields, for example: Give me all samples/data files HCD or with this particular PTM or for this particular instrument. This information can help a lot with the querying and descriptions at the dataset/sample level.

It's entirely reasonable to include "which dissociation methods does this dataset use" in SDRF. It's just not really sensible to tie it with ms level as if that was really sufficient to say which dissociation method each scan used. Likewise, mzML includes which spectra types are in a file in the fileContent element, but it doesn't tie it to ms level. SDRF should/could also do that so it's easy to query what kind of spectra are in a file.

PR #222, opened 2023-11-03
Add first batch of mzqc readers/writers

cbielow:
@mwalzer 
@nilshoffmann
@bittremieux 
please have a look. Thanks! :)

I used the MS:4....... space, since its QC software.

Was also wondering if we should have a new term like
```
[Term]
id: MS:?????
name: quality control software
def: "Quality Control software." [PSI:MS]
is_a: MS:1000531 ! software
```
which we can derive from.

### DIFF
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -24041,6 +24041,24 @@ synonym: "MS2-PrecZ-5" RELATED [PMID:24494671]
 synonym: "MS2-PrecZ-more" RELATED [PMID:24494671]
 
 [Term]
+id: MS:4000177
+name: rmzqc
+def: "An R package for reading, validating, and writing mzQC files." [https://github.com/MS-Quality-hub/rmzqc]
+is_a: MS:1001457 ! data processing software
+
+[Term]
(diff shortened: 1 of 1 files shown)


edeutsch:
I'm thinking software package names ought to go in the MS:1000 series. What do you think?

cbielow:
Fine with me as well.

nilshoffmann:
@cbielow Should we rather use
`is_a: MS:1001457 ! data processing software` for the new root term "Quality Control software"?

cbielow:
ok, done.

@bittremieux  @mwalzer Is this good to go?

edeutsch (approved):

edeutsch:
We looked at the latest diff at the PSI-MS and were all happy with this as is. Waiting for review by @bittremieux and @mwalzer before merge.

mobiusklein (approved):

bittremieux (approved):

cbielow:
We've briefly touched upon this PR in the mzQC meeting yesterday, so we should be good to merge.

PR #230, opened 2023-12-15
Add ion mobility attribute as parent for CCS

chambm:
As discussed on PSI call, we want to allow CCS to be encoded in mzML wherever individual ion mobility values are currently allowed.

### DIFF
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -1,5 +1,5 @@
 format-version: 1.2
-data-version: 4.1.137
+data-version: 4.1.138
 date: 07:12:2023 16:37
 saved-by: Matt Chambers
 auto-generated-by: OBO-Edit 2.3.1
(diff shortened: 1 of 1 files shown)


edeutsch (approved):

edeutsch:
This looks good, thanks. It was mentioned at the call today that several vendor APIs allow for computing the CCS when reading vendor datafiles, so this could be easily encoded in the mzMLs.

mobiusklein (approved):
Looks good to me. There's a degree of approximation error that we don't really understand but there's nothing to be done about it. Because CCS requires knowledge of charge state, it doesn't yet seem feasible to encode on a peak level.

edeutsch:
but this is for precursors I think, right? where the charge state is ostensibly known?

chambm:
Yes, it only makes sense when charge state is relevant. So unlike drift time et al., it actually doesn't make sense in `<scan>`, only in `<selectedIon>`. However, it is feasible to have an array of CCS values just like we can have an array of charge states. Unknown values could be stored as 0. I think we can leave that for another PR though.

PR #232, opened 2024-01-24
Map MS partof to RO

matentzn:
Fixes https://github.com/EBISPOT/ols4/issues/600

Ideally, in a future PR, you should consider to import RO rather than copy pasting the semantics here (i.e. transitivity).

### DIFF
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -45,6 +45,7 @@ name: has_units
 [Typedef]
 id: part_of
 name: part_of
+xref: BFO:0000050
 is_transitive: true
 
 [Typedef]
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -1,7 +1,7 @@
 format-version: 1.2
-data-version: 4.1.138
-date: 07:12:2023 16:37
-saved-by: Matt Chambers
+data-version: 4.1.139
+date: 24:01:2024 10:23
+saved-by: Yasset Perez-Riverol
 auto-generated-by: OBO-Edit 2.3.1
 import: http://purl.obolibrary.org/obo/pato.obo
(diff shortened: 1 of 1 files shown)


ypriverol (approved):

psi-ms.obo
```diff
@@ -45,6 +45,7 @@ name: has_units
 [Typedef]
 id: part_of
 name: part_of
+xref: BFO:0000050
```
ypriverol, line 48: @matentzn Can you update the version of the ontology. See the failing Actions https://github.com/HUPO-PSI/psi-ms-CV/actions/runs/7638645173/job/20810093289

matentzn: Feel free to close this PR and redo it, I just wanted to show you how you can map your part of to RO, so it displays correctly in OLS and interoperates with other ontologies. You can also directly edit my PR and make the required changes!

ypriverol: Done. Thanks a lot for the tip.

edeutsch (approved):
It would be funny if after all this pain, the solution is as simple as this..

ypriverol:
I merged the PR to see the changes and opened the owl created in Protege as suggested; and it looks for me that the problem still remains 

Screenshot 2024-01-24 at 20 17 51

Any advice @matentzn ?

matentzn:
This is a weakness in Protege I think. If you look at the peptide classes, you can see peptide attribute underneath, with the blue arrow..

image

Lets see how Protege will display this..

matentzn:
(I think the problem in Protege is that it will include the is a hierarchy no matter what, and superimposes the other "relation ships" after the fact - which means that what you see in terms of top level roots are all the roots according to the is a hierarchy)

PR #235, opened 2024-02-09
Update SMILES formula to SMILES string

edeutsch:


mobiusklein (approved):

edeutsch:
Anything need to be done before merging this? I think we agreed at the call to merge, so don't need any more approvals, I think.

PR #245, opened 2024-03-17
Add missing value types to mzSpecLib attributes

mobiusklein:


edeutsch (approved):
This is good and as was originally intended.
Need to update version information based on recent commits

PR #246, opened 2024-03-19
Protein contaminant ratio

cbielow:
Note: new version assumes that https://github.com/HUPO-PSI/psi-ms-CV/pull/245 is merged first.

### DIFF
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -24060,6 +24060,19 @@ synonym: "MS2-PrecZ-5" RELATED [PMID:24494671]
 synonym: "MS2-PrecZ-more" RELATED [PMID:24494671]
 
 [Term]
+id: MS:4000177
+name: protein contaminant intensity ratio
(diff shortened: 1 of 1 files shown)


bittremieux (approved):

cbielow:
IMHO ready for merge from our side :)

edeutsch (commented):
This looks okay to me, except maybe the "intensity" in the name. The name has "intensity" but the description only talks about "abundance".  These two concepts are related but not really equal. "ratio" and "fraction" are also variously used in ways that are unclear. "fraction of total protein" is clear. "ratio" only serves to confuse the issue, I think. 

I'm thinking the name would be better "contaminant protein abundance fraction".

cbielow:
point well made.
Updated the name accordingly.

edeutsch (approved):
I still think it is better as "contaminant protein abundance fraction", i.e. with contaminant first, because we are talking about "contaminant protein"s and not really "protein contaminant"s.

But adequate as is.

cbielow:
whops. Fixed the name. Thanks Eric.
Also updated the `data-version`, so it can be merged directly without relying on  #245 

Ready to merge.

PR #247, opened 2024-03-20
De novo database format work around

mobiusklein:
As discussed during the spring meeting.


edeutsch (commented):
Can we just change the name to "no sequence database"

edeutsch (commented):
def: "No reference sequence database was used in the search process to determine the identified peptide sequence, for example as with de novo sequencing."

edeutsch (approved):

PR #248, opened 2024-03-21
Add CV term for mzRecal software

524D:
Hi,

I would like to add a new CV term for mzRecal. The PR should have all the info, the CV number still needs to be filled in though.
Please let me know if this is sufficient.

Rob


edeutsch (commented):
Looks pretty good. A few changes:
- Remove m/z calibration: software should not be child of protocol
- Add latest number
- Add version change
-

PR #249, opened 2024-04-04
Update psi-ms.obo

deeptijk:
Adding a term for new instrument model Bruker timsTOF HT


edeutsch (approved):

ypriverol (changes requested):
@deeptijk @deeptijk you have to increase manually the version of the file and also add your user as the last person that change the file.

PR #254, opened 2024-04-08
add precursor ppm accuracy and precision

### LINKED ISSUE #241: QC metrics to track (absolute) delta ppm errors for identification
### What is the QC term's name?

absolute dppm sigma and absolute dppm mean

### Briefly describe the QC term.

The terms track the mean and sigma of the error distributions of spectra identification. For the proposed definition, please see below.

### What is the QC term's unit?

_No response_

### Value type

MS:4000003 ! single value

### Describe any additional information.

Here are drafts for the terms' entries
```
[Term]
id: MS:4000xxx
name: absolute dppm mean
def: "From the distribution of absolute values of observed mass accuracies (MS:4000072) in ppm of identified spectra within the run after user defined acceptance criteria are applied, the mean" [PSI:MS]
is_a: MS:4000003 ! single value
relationship: has_metric_category MS:4000008 ! ID based metric
relationship: has_metric_category MS:4000021 ! MS1 metric
relationship: has_value_concept STATO:0000401 ! sample mean
relationship: has_value_concept MS:1000014 ! accuracy
relationship: has_units UO:0000169 ! parts per million
relationship: has_value_type xsd:float ! The allowed value-type for this CV term

[Term]
id: MS:4000xxx
name: absolute dppm sigma
def: "From the distribution of absolute values of observed mass accuracies (MS:4000072) in ppm of identified spectra within the run after user defined acceptance criteria are applied, the sigma value" [PSI:MS]
is_a: MS:4000003 ! single value
relationship: has_metric_category MS:4000008 ! ID based metric
relationship: has_metric_category MS:4000021 ! MS1 metric
relationship: has_value_concept STATO:0000237 ! standard deviation
relationship: has_units UO:0000169 ! parts per million
relationship: has_value_concept MS:1000014 ! accuracy
relationship: has_value_type xsd:float ! The allowed value-type for this CV term
```

cbielow:
fixes  #241

I've opted to add a `precision` term, which I'm not sure is needed, but since there is `accuracy` already, I thought there is no harm.
However, those two terms are rather narrow, but maybe that's ok. I used a free ID slot (MS:1000394).
We can also remove the precision annotation if desired. I have no strong opinion, but I think may add some helpful context.

### DIFF
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -24099,7 +24099,7 @@ synonym: "PG: Contaminants" NARROW []
 [Term]
 id: MS:4000178
 name: mean of precursor ppm deviation
-def: "The mean of observed precursor mass accuracies (MS:4000072) in ppm of identified MS2 spectra after user-defined acceptance criteria (FDR) are applied" [PSI:MS]
+def: "The mean of the distribution of observed precursor mass accuracies (MS:4000072) [in ppm] of identified MS2 spectra after user-defined acceptance criteria (FDR) are applied" [PSI:MS]
 is_a: MS:4000003 ! single value
 relationship: has_metric_category MS:4000008 ! ID based metric
 relationship: has_metric_category MS:4000021 ! MS1 metric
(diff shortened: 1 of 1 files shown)


mwalzer:
Maybe align the name wording to the other distribution parameter QC metric terms? 
e.g. MS:4000108 and [MS:4000109](https://github.com/HUPO-PSI/psi-ms-CV/blob/24dfb9733249e1814cee79cfd55642233b28c543/psi-ms.obo#L23321)

mwalzer:
:+1:

edeutsch (commented):
Our custom is to have names like:
  precursor ppm deviation mean
  precursor ppm deviation standard deviation
instead of terms that include "of"

bittremieux (changes requested):
@cbielow Can you make the naming changes that Eric requested? Other than that it looks good to me.

bittremieux (approved):

psi-ms.obo
```diff
@@ -2895,9 +2895,19 @@ is_a: MS:1000247 ! desorption ionization
 
 [Term]
 id: MS:1000394
+<<<<<<< HEAD
```
bittremieux, line 2898: Can you fix the merge conflicts here?

bittremieux (changes requested):

cbielow:
should be good to go now

bittremieux (approved):

edeutsch (approved):

Issue #243, opened 2024-02-23
A QC metric to track the duration it takes to 'accumulate' each subsequent quarter of identified spectra


mwalzer:
### What is the QC term's name?

identified MS2 quarter RT fraction

### Briefly describe the QC term.

The interval in which the first, second, third, and fourth quarter of all identified MS2 events are, divided by retention time duration. The metric informs about the dynamic range of the MS2 identification process along the chromatographic separation. For MS2 scans, the comparability of the values depends on the acquisition mode and settings to select ions for fragmentation.

### What is the QC term's unit?

It's a fraction (s/s)

### Value type

MS:4000004 ! n-tuple

### Describe any additional information.

Term entry proposition:
```
[Term]
id: MS:4000xxx
name: identified MS2 quarter RT fraction
def: "The interval in which the first, second, third, and fourth quarter of all identified MS2 events are, divided by retention time duration." [PSI:MS]
comment: The metric informs about the dynamic range of the MS2 identification process along the chromatographic separation. For MS2 scans, the comparability of the values depends on the acquisition mode and settings to select ions for fragmentation.
is_a: MS:4000004 ! n-tuple
relationship: has_metric_category MS:4000008 ! ID based metric
relationship: has_metric_category MS:4000012 ! single run based metric
relationship: has_metric_category MS:4000016 ! retention time metric
relationship: has_value_type xsd:float ! The allowed value-type for this CV term
relationship: has_units UO:0000191 ! fraction
```

cbielow:
maybe add 
```
synonym: "RT-MS-Q1" RELATED [PMID:24494671]
synonym: "RT-MS-Q2" RELATED [PMID:24494671]
synonym: "RT-MS-Q3" RELATED [PMID:24494671]
synonym: "RT-MS-Q4" RELATED [PMID:24494671]
```
to give some hint on what exactly these 4 quarters are? But maybe even better, we should be more concise in the definition, so people to not need to read a paper to understand what is meant here?

cbielow:
as discussed, this wording may need a little tinkering, i.e.
given [p0, p25, p50,p75,p100] as percentiles of the above MS2 event distributions in unit of seconds, what this actually computes is
{ (p25 - p0) / (p100 - p0),
  (p50 - p25) / (p100 - p0),
 ....}

even though I'm not sure if 'RT duration' is its total duration or just the span in which IDs are observed. The implications are quite different.

mwalzer:
[...]

Issue #242, opened 2024-02-23
A QC metric to track info surrounding enzymatic digestion


mwalzer:
### What is the QC term's name?

enzyme digestion parameters

### Briefly describe the QC term.

This metric is designed to track data for identified spectra, hence a tabular design, for now tracking the spectrum number and the number of missed cleavages. 

### Value type

MS:4000005 ! table

### Describe any additional information.

Here the term's proposed entry:
```
[Term]
id: MS:4000xxx
name: enzyme digestion parameters
def: "Tabular representation of the parameters around enzymatic digestion." [PSI:MS]
is_a: MS:4000005 ! table
relationship: has_metric_category MS:4000008 ! ID based metric
relationship: has_metric_category MS:4000012 ! single run based metric
relationship: has_column MS:1000767 ! native spectrum identifier format
relationship: has_column MS:1003044 ! number of missed cleavages
relationship: has_optional_column MS:1000511! ms level
```

mwalzer:
@cbielow @bittremieux  Any comments before I'll create a PR?

mwalzer:
digest from slack:
```
[Term]
id: MS:4000xxx
name: missed cleavages histogram
def: "The number of missed cleavages, tallied by the number of occurrence of n missed cleavages per spectrum for a set of identified MS2 spectra after user-defined acceptance criteria (FDR) are applied. The number of missed cleavages per spectrum is given in one column, whereas the occurrence column gives the counts of spectra with the according number of missed cleavages. The highest missed cleavage state is to be interpreted as that number of missed cleavages or more." [PSI:MS]
is_a: MS:4000005 ! table
relationship: has_metric_category MS:4000008 ! ID based metric
relationship: has_metric_category MS:4000022 ! MS2 metric
relationship: has_column MS:1003044 ! number of missed cleavages
relationship: has_column NCIT:C150827 ! Number of Occurrences
```
tallied <-> binned ?
Is the second sentence of the definition necessary?
Are there any synonymous metrics I missed?

bittremieux:
Completed in #255.

PR #255, opened 2024-04-12
add terms from #242 and #243

mwalzer:
also added dependency NCIT terms

### DIFF
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -186,6 +186,18 @@ def: "One of the axes representing the projection of varience resulting from pri
 is_a: NCIT:C19044 ! Statistical Technique
 
 [Term]
+id: NCIT:C150827
+name: Number
+def: "A numeral or string of numerals expressing value, quantity, or identification." [] {http://purl.obolibrary.org/obo/NCIT_C150827="NCI"}
+is_a: NCIT:C20189 ! Property or Attribute
+
+[Term]
+id: NCIT:C150827
+name: Number of Occurrences
+def: "The number of times something happened." [] {http://purl.obolibrary.org/obo/NCIT_C150827="NCI"}
+is_a: NCIT:C150827 ! Number
+
+[Term]
 id: MS:0000000
 name: Proteomics Standards Initiative Mass Spectrometry Vocabularies
 def: "Proteomics Standards Initiative Mass Spectrometry Vocabularies." [PSI:MS]
@@ -24085,6 +24097,33 @@ relationship: has_value_type xsd:float ! The allowed value-type for this CV
 relationship: has_domain MS:1002305 ! value between 0 and 1 inclusive
 relationship: has_order MS:1002109 ! lower score better
 synonym: "PG: Contaminants" NARROW []
+
+[Term]
+id: MS:4000180
+name: missed cleavages histogram
+def: "The number of missed cleavages, tallied by the number of occurrence of n missed cleavages per spectrum for a set of identified MS2 spectra after user-defined acceptance criteria (FDR) are applied. The number of missed cleavages per spectrum is given in one column, whereas the occurrence column gives the counts of spectra with the according number of missed cleavages. The highest missed cleavage state is to be interpreted as that number of missed cleavages or more." [PSI:MS]
+is_a: MS:4000005 ! table
+relationship: has_metric_category MS:4000008 ! ID based metric
+relationship: has_metric_category MS:4000022 ! MS2 metric
+relationship: has_column MS:1003044 ! number of missed cleavages
+relationship: has_column NCIT:C150827 ! Number of Occurrences
+
+[Term]
+id: MS:4000181
+name: identified MS2 quarter RT fraction
+def: "The retention time interval in which the first, second, third, and fourth quarter of all identified MS2 events, ordered by retention time, are recorded, each divided by total retention time duration." [PSI:MS]
+comment: The metric informs about the dynamic range of the MS2 identification process along the chromatographic separation. For MS2 scans, the comparability of the values depends on the acquisition mode and settings to select ions for fragmentation.
+is_a: MS:4000004 ! n-tuple
+relationship: has_metric_category MS:4000012 ! single run based metric
+relationship: has_metric_category MS:4000008 ! ID based metric
+relationship: has_metric_category MS:4000022 ! MS2 metric
+relationship: has_metric_category MS:4000016 ! retention time metric
+relationship: has_value_type xsd:float ! The allowed value-type for this CV term
+relationship: has_units UO:0000191 ! fraction
+synonym: "RT-MS-Q1" RELATED [PMID:24494671]
+synonym: "RT-MS-Q2" RELATED [PMID:24494671]
+synonym: "RT-MS-Q3" RELATED [PMID:24494671]
+synonym: "RT-MS-Q4" RELATED [PMID:24494671]
 
 [Term]
 id: PEFF:0000001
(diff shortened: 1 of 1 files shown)


psi-ms.obo
```diff
@@ -185,6 +185,18 @@ name: Principal Component
 def: "One of the axes representing the projection of varience resulting from principal component analysis." [] {http://purl.obolibrary.org/obo/NCIT_C60694="NCI"}
 is_a: NCIT:C19044 ! Statistical Technique
 
+[Term]
+id: NCIT:C150827
```
mobiusklein, line 189: This `id` is incorrect, referring to the same "Number of Occurrences" term.

mobiusklein (changes requested):

psi-ms.obo
```diff
@@ -24086,6 +24098,33 @@ relationship: has_domain MS:1002305 ! value between 0 and 1 inclusive
 relationship: has_order MS:1002109 ! lower score better
 synonym: "PG: Contaminants" NARROW []
 
+[Term]
+id: MS:4000180
+name: missed cleavages histogram
```
bittremieux, line 24103: I would rename this to "number of missed cleavages" instead of "histogram".

psi-ms.obo
```diff
@@ -24086,6 +24098,33 @@ relationship: has_domain MS:1002305 ! value between 0 and 1 inclusive
...
+def: "The number of missed cleavages, tallied by the number of occurrence of n missed cleavages per spectrum for a set of identified MS2 spectra after user-defined acceptance criteria (FDR) are applied. The number of missed cleavages per spectrum is given in one column, whereas the occurrence column gives the counts of spectra with the according number of missed cleavages. The highest missed cleavage state is to be interpreted as that number of missed cleavages or more." [PSI:MS]
```
bittremieux, line 24104: "per spectrum" -> "per PSM" to also cover multiple assignments from chimeric spectra.

bittremieux, line 24104: In general I think this definition is more complex than it should be, and makes it challenging to understand what the metric should look like.

See for analogy the similar precursor charge fractions, which is much easier to understand imo:
> def: "The fraction of MS/MS precursors of the corresponding charge. The fractions [0,1] are given in the 'Fraction' column, corresponding charges in the 'Charge state' column. The highest charge state is to be interpreted as that charge state or higher." [PSI:MS]

psi-ms.obo
```diff
@@ -24086,6 +24098,33 @@ relationship: has_domain MS:1002305 ! value between 0 and 1 inclusive
...
+is_a: MS:4000005 ! table
+relationship: has_metric_category MS:4000008 ! ID based metric
+relationship: has_metric_category MS:4000022 ! MS2 metric
+relationship: has_column MS:1003044 ! number of missed cleavages
+relationship: has_column NCIT:C150827 ! Number of Occurrences
```
bittremieux, line 24109: Should this be a number or a percentage? If the total number of identified spectra is included, they can easily be interconverted. But as an independent metric, I think the fraction is more informative, so I'd lean towards that instead.

psi-ms.obo
```diff
@@ -24086,6 +24098,33 @@ relationship: has_domain MS:1002305 ! value between 0 and 1 inclusive
...
+
+[Term]
+id: MS:4000181
+name: identified MS2 quarter RT fraction
+def: "The retention time interval in which the first, second, third, and fourth quarter of all identified MS2 events, ordered by retention time, are recorded, each divided by total retention time duration." [PSI:MS]
```
bittremieux, line 24114: Simplify the definition in analogy to related terms. E.g.:
> def: "The interval used for acquisition of the first, second, third, and fourth quarter of all MS2 events divided by retention time duration." [PSI:MS]

Basically the only thing that needs to be added is "identified" to MS2.

psi-ms.obo
```diff
@@ -24086,6 +24098,33 @@ relationship: has_domain MS:1002305 ! value between 0 and 1 inclusive
...
+comment: The metric informs about the dynamic range of the MS2 identification process along the chromatographic separation. For MS2 scans, the comparability of the values depends on the acquisition mode and settings to select ions for fragmentation.
+is_a: MS:4000004 ! n-tuple
+relationship: has_metric_category MS:4000012 ! single run based metric
+relationship: has_metric_category MS:4000008 ! ID based metric
+relationship: has_metric_category MS:4000022 ! MS2 metric
+relationship: has_metric_category MS:4000016 ! retention time metric
+relationship: has_value_type xsd:float ! The allowed value-type for this CV term
+relationship: has_units UO:0000191 ! fraction
+synonym: "RT-MS-Q1" RELATED [PMID:24494671]
```
bittremieux, line 24123: The closer synonyms are the "MSMS" versions of these metrics instead of the "MS" versions.

bittremieux (changes requested):

psi-ms.obo
mwalzer: Fixed

psi-ms.obo
mwalzer: see updates

psi-ms.obo
mwalzer: see updates

psi-ms.obo
mwalzer: see updates

psi-ms.obo
mwalzer: fixed

psi-ms.obo
```diff
@@ -24092,6 +24105,33 @@ relationship: has_domain MS:1002305 ! value between 0 and 1 inclusive
 relationship: has_order MS:1002109 ! lower score better
 synonym: "PG: Contaminants" NARROW []
 
+[Term]
+id: MS:4000180
+name: table of missed cleavage counts
+def: "The number of identified PSM with corresponding number of missed cleavages. The number of missed cleavages per PSM is given in the 'number of missed cleavages' column, the respective count of such PSM in the 'Number of Occurrences' column. The highest 'missed cleavages' row is to be interpreted as that number of missed cleavages or higher." [PSI:MS]
+is_a: MS:4000005 ! table
+relationship: has_metric_category MS:4000008 ! ID based metric
+relationship: has_metric_category MS:4000022 ! MS2 metric
```
bittremieux, line 24114: Somewhat related to the question above. If we're talking about peptides rather than PSMs, is this truly an MS2-level metric? The peptide sequences themselves are not necessarily exclusively derived from MS2 data, e.g. identifications from MS1 during match-between-runs.

psi-ms.obo
```diff
@@ -24092,6 +24105,33 @@ relationship: has_domain MS:1002305 ! value between 0 and 1 inclusive
 relationship: has_order MS:1002109 ! lower score better
 synonym: "PG: Contaminants" NARROW []
 
+[Term]
+id: MS:4000180
+name: table of missed cleavage counts
+def: "The number of identified PSM with corresponding number of missed cleavages. The number of missed cleavages per PSM is given in the 'number of missed cleavages' column, the respective count of such PSM in the 'Number of Occurrences' column. The highest 'missed cleavages' row is to be interpreted as that number of missed cleavages or higher." [PSI:MS]
```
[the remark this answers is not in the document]
bittremieux: I agree, the metric relates to the digestion step, so whether a peptide is sampled multiple times in the MS doesn't really matter.

I think "peptides" feels a bit more natural than the very generic "sequences".

psi-ms.obo
mwalzer: To the effect of the discussion above, idk about the MS1, but I think it's certainly divorced from the MS2 concept. Just remove the MS2 metric category?

bittremieux: Yes, I agree.

mobiusklein (approved):

PR #273, opened 2024-05-09
Scout software added to the ontology

ypriverol:
- Scout software added
- Scout score added


diogobor (approved):
Scout ontologies are correct.

mobiusklein (approved):

edeutsch (commented):
Is there a citation that should be referenced in the square brackets?

ypriverol:
Paper under review, I will add the reference after publication out.

PR #278, opened 2024-06-03
Add Thermo Stellar MS

edeutsch:


chambm:
I was wondering whether it should mention that the ion trap is dual pressure, but then I remember the Velos is as well, and it seems the older Thermo CV terms don't include instrument components at all. So what's there is probably fine.

chambm (approved):

jshofstahl:
Looks good to me.

mobiusklein (approved):
After fixing the term ID, this looks fine to me.

vagisha (approved):
Looks good to me.  Thank you.

PR #288, opened 2024-07-03
add CV term for electron beam energy used in EAD

david-cox-sciex:
New CV term for electron beam energy. This is the kinetic energy of the electron beam using in dissociation methods induced by a free electron beam, such as electron-capture dissociation (ECD), electron-detachment dissociation (EDD), and electron-activated dissociation (EAD).


psi-ms.obo
```diff
@@ -22503,6 +22503,15 @@ name: Stellar
 def: "Thermo Scientific Stellar mass spectrometer contains a quadrupole mass filter, a collision cell, and a quadrupole linear ion trap mass analyzer." [PSI:MS]
 is_a: MS:1000494 ! Thermo Scientific instrument model
 
+[Term]
+id: MS:1003410
+name: electron beam energy
+def: "The kinetic energy of the electron beam using in dissociation methods induced by a free electron beam, such as electron-capture dissociation (ECD), electron-detachment dissociation (EDD), and electron-activated dissociation (EAD)." [PSI:MS]
```
chambm, line 22509: Typo here: "using in" -> "used in"

mobiusklein:
The term itself looks good to me. I don't know enough mechanistically to say whether this would apply to ETD too though.

Ontologically, should we make a new parent term for all activation energies? We have `activation energy`, `collision energy`, `normalized collision energy`, and the related ramp terms as well as `supplemental collision energy`. This adds yet another term of the same concept but different mechanism.

mobiusklein (approved):

david-cox-sciex:
A parent term for all activation energy terms seems good.
I considered using "activation energy" but since "electron beam energy" is about the energy the electron has, it seemed wrong to use a term that is about the final precursors activation energy. Also, when the spectrum is the result of mixed CID and EAD process we would need both "electron beam energy" and "collision energy" terms.

chambm:
@mobiusklein Is this ready to merge or do we need to wait for the OLS stuff to settle down?

david-cox-sciex:
I'm getting an error about it not being able to generate-owl. This branch has one new term and I incremented the data version count be one. I'm not sure why it is failing.

mobiusklein:
We introduced a new workflow to fix an issue with Ontology Lookup Service. We will likely merge this tomorrow.

edeutsch (approved):

PR #295, opened 2024-07-12
Validation of the OWL file for OLS

ypriverol:
@mobiusklein @edeutsch : 

I have tried to get an action workflow that validates the OWL. Unfortunately, the code they have downloads the files from the given path and does not allow to pass files for validation. Then, we will need to validate existing PRs, unless we release the OBO file first into a branch and then validate there and PR into master. The current version of the OWL file is still failing BTW. The workflow does the following: 

- Download the dataload tool from OLS4 which is the tool that loads the ontologies. 
- Build the tool. 
- Download a config file that only contains our ontology to be validated. 
- Perform the conversion/validation from RDF2JSON (really don't understand what they are trying to achieve, but is where we are failing). 
- Compare the output file with a given file with validation and empty file, if they are equal, then we are not passing "validation"


mobiusklein:
Thank you for setting this up. It looks like it is doing what we need it to, I have to get to work, but will come back to this when I have time later today.

ypriverol:
@mobiusklein I think the validation is ready. You can see here how the validation works. With the current OWL, it works.

mobiusklein (approved):
I made a slight adjustment to the `README` but this looks good to me.

PR #296, opened 2024-07-12
Stato & PATO fixes

ypriverol:
All STATO relationships for QC terms are removed. They were in total 54 relationships which is difficult to copy manually if you also need to add parent terms. Then, I just removed the relationship and the import.


mobiusklein (approved):
In the interest of immediately resolving these issues, let's run with this. It passes through OWL conversion successfully.

@bittremieux we will need you or someone else from MZQC to work with us to find a suitable replacement for those removed concepts, which we can retrieve from the previous release/tag.

ypriverol:
I also vote to remove all the references to the following terms: 
```
relationship: has_units UO:00000XX

edeutsch (commented):
If this is an emergency fix, I guess it is okay, but I think these relationships are needed. If we merge, I think these will need to be restored in the future. Is that the plan?

ypriverol:
> If this is an emergency fix, I guess it is okay, but I think these relationships are needed. If we merge, I think these will need to be restored in the future. Is that the plan?

We can try to recover them in the future, However, we do have also other terms that have perfect relations with other terms in external ontologies and we don't add the relationship.

ypriverol:
@edeutsch can you let me know if the PR is OK Approve or Reject? Sorry for the pressure.

edeutsch (approved):
okay, I'll go ahead and approve it so that we can get the flakey OLS running again, but I suspect our mzQC friends will not be content with this change and another solution will be needed. Tagging @bittremieux who maybe get help out with a solution. I suspect that this commit will break mzQC related software.

bittremieux:
Sorry that I couldn't join the meeting yesterday—it was a holiday here.

This is quite problematic for mzQC, because we use these terms for semantic validation. What is the problem with these terms vs several other terms coming from external ontologies? Why can't STATO be used?

mobiusklein:
STATO apparently changed something about their OWL file such that OLS4 rejected it, but OLS4 wasn't specific about why.

Remediation options are:
  1. Wait for STATO to fix things, restore the imports and then use Yasset's new validation workflow as an early warning system prior to releasing new versions. This is undesirable because it still leaves us vulnerable to being unlisted because STATO breaking still leads to us breaking. I suspect this is even worse than I perceive it to be because by the time I wake up and see these alerts, Yasset has already had to deal with the PRIDE service going down for most of the day.
  2. Inline the terms from STATO into our CV like we did for UO and NCIT. This works, but we could end up porting more than the 6-7 terms we actually use in order to get to complete the subgraph of STATO actually in use.
  3. Define new terms in MS that replace the STATO terms, mark them as synonyms with the STATO terms, and then rewrite your validator to look for our synonyms instead.

ypriverol:
I want to make clear that not only PRIDE rely on OLS. A couple of months ago, I was @lgatto the one that reported the issue https://github.com/EBISPOT/ols4/issues/600#issuecomment-2079323403 . I mentioned yesterday in the meeting @bittremieux, that dependencies from other ontologies should be carefully introduced in PSI MS because if the ontology is failing, you have to contact external maintainers to get support. @mobiusklein mentioned that some ontologies haven't been updated in 3 years, so you don't have anyone to help in case something happens.

bittremieux:
That makes sense. However, in this case STATO was actually updated [mere days ago](https://github.com/ISA-tools/stato/commits/dev/). So it seems like they'd be responsive in helping getting stuff fixed. And I don't think it makes sense to have to define a bunch of out-of-context terms ourselves. That's pretty much directly violating the whole point of ontologies.

ypriverol:
Looks like dependencies are complex right now. Because you can't control the import version of other ontologies. This means if the dependency (import) fails in OLS in the next release your ontology will again fail. We really need to evaluate the import and just justify the import if is needed. Even if the STATO is "actively" maintained (let's say they respond within 5 days), the level of updates that PSI-MS have right now (almost daily) and the dependencies in the production environments ProteomeXchange and other PSI tools will make this complex. I think this should be discussed pros and cons and see how many terms we need to add from STATO.

mobiusklein:
The root of the problem isn't just dependencies. It's that OLS is an abstraction over the distinct CV/ontologies, and that it has leaked into all of its client applications. When it fails, everything falls apart with no fallback mechanism. This makes its clients sensitive to fluctuations in maintenance across the entire dependency graph of the CVs they use.

If we ignore OLS, we would use pURLs with or without version pinning and then it's an implementation detail for the application to resolve, parse, and merge the term graphs, and handle fallbacks for when remote sources aren't available. That can be a lot of work. I think that work is worthwhile if the cost of OLS taking down the CV is large. I do not know what is involved in doing that for PRIDE though.

ypriverol:
@mobiusklein not only PRIDE use OLS, but other proteomics tools, apart from users that use the interface to search for terms. Let's imagine for now that PRIDE move away from PSI-MS or implements a mechanism to have local versions of the ontologies (both solutions possible). 

You will end up in the same situation every time the ontology gets removed from OLS because of errors in dependencies and we will need to take the responsibility to contact all the developers of our dependencies. My entire point is that `importing` poses a challenge to the ontology to follow up for errors in the corresponding dependencies. If we are willing to do that, Im also fine with it, and we in PRIDE will need to find a solution for the dependency from OLS.

edeutsch:
I've always wondered if referencing other ontologies without importing them is a permissable mode? I seem to recall a discussion with an OLS person who recently said that was fine. Maybe we can just skip the import statements and it would be fine/better?

edeutsch:
Maybe another option is to repeat all foreign term definitions in our file like with did with NCIT. And then also not import anything.

cmungall:
I encourage this community to make use of OBO standards here. There are well-known and understood issues related to ontology modularity.

It seems that you removed potentially valuable inter-ontology linkages because of some temporary issue with an upstream OWL file. These kinds of issues should be avoided by switching imports to a local module.

> I've always wondered if referencing other ontologies without importing them is a permissable mode?

There is a standard product called a base module which is what you are after here.

proccaserra:
@cmungall it was at the time of me updating the information regarding STATO. 
If PSI-MS have changed their practice, I'll update the record accordingly.
I guess there are enough users to pass the criteria of plurality. 
All issues with STATO have been addressed and is now served from OLS again.
Thanks @bittremieux @mobiusklein @ypriverol @edeutsch

mobiusklein:
@cmungall by "base module", are you referring to different subsets of an ontology, e.g. `go-slim` vs `go`, or is this another mechanism? Is there an example you can point me to about how this is done?

Does the term piracy cause a problem for consumers of larger term-graphs, or is it more a long-term maintenance burden issue?

PR #299, opened 2024-07-18
Restore deleted STATO references to evaluate validation without import

mobiusklein:
This restores the references to `STATO` removed in #296 without actually importing it.

It appears to work by just adding nodes for these classes' IDs, but provides zero definition for them, so it would not know about parents or other higher order relations depending upon the `STATO` terms, but still be able to say "these are sample means and these are quantiles"

@bittremieux are these sufficient for MzQC's validator?

### DIFF
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -22672,6 +22672,8 @@ relationship: has_metric_category MS:4000012 ! single run based metric
 relationship: has_metric_category MS:4000018 ! XIC metric
 relationship: has_value_type xsd:float ! The allowed value-type for this CV term
 relationship: has_value_concept MS:1000086 ! full width at half-maximum
+relationship: has_value_concept STATO:0000291 ! quantile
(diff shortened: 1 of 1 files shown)


edeutsch (approved):
I think this is probably a good way to go. It might cause some breakage somewhere, but if it makes the very delicate OLS happy, then I'll support it.

bittremieux (approved):
After discussion during the QC call, this solution satisfies our requirements.

bittremieux:
Side note: the PATO terms that were deleted in #296 have not been restored here. These are not used by the QC terms, but rather by the copied UO terms at the end of the CV.

This also raises a discrepancy between how terms from external CVs/ontologies are used: STATO terms are now used without a reference to their source ontology, some UO terms have been redefined in our CV, and there might be some additional differences as well. Should we try to harmonize these approaches for external terms, also in a way that keeps OLS happy?

bittremieux (approved):

PR #303, opened 2024-07-24
Clarify protein contaminants term

bittremieux:
- Be more precise that it's a ratio rather than a general fraction.
- Add the ratio unit.


edeutsch (commented):
I thought we previously changed this to "fraction" at initial creation? A ratio is defined as a relation between two numbers, but as a datatype float, there can only be one number, and as it is usually a single non-whole number, it is thus a fraction, right?
"150:1500" or "1:10" would be a ratio. "0.1" would be the fraction?

edeutsch (approved):

PR #316, opened 2024-08-05
Harmonize fraction as value concept

bittremieux:
Fixes #307.

PR #318, opened 2024-08-07
Add counts units

bittremieux:
Add counts units to terms that are part of the MS1XXXXX namespace but can be used as QC-related terms as well. This is necessary for compliant mzQC files and semantic validation.


edeutsch (approved):
Looks good. Needs version bump after #316 is merged.

mobiusklein (approved):

PR #330, opened 2024-09-16
Accession order fixes

cbielow:
Enhances `scripts/find_next_available_number.py` to report the gap size and also now detects an unsorted CV (which would cause reporting of false positive gaps).

e.g.
```
Found gap of size 62: ('MS', 1000938) to ('MS', 1000999)
...
Found gap of size 1: ('MS', 1003236) to ('MS', 1003236)
Found gap of size 2: ('MS', 1003407) to ('MS', 1003408)               <-- false positive...
Found gap of size 2996587: ('MS', 1003413) to ('MS', 3999999)
Found gap of size 1: ('MS', 4000011) to ('MS', 4000011)
Found gap of size 25: ('MS', 4000025) to ('MS', 4000049)
CV is not sorted! MS:1003407 found after MS:4000181
Found gap of size 1000965: ('PEFF', 1036) to ('PEFF', 1002000)
```

Ordering of accessions is restored as well in this PR (eliminating `CV is not sorted! MS:1003407 found after MS:4000181`).

This might be worth adding to GH actions? i.e. fail if not sorted?

### DIFF
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -1,6 +1,6 @@
 format-version: 1.2
-data-version: 4.1.176
-date: 17:09:2024 07:15
+data-version: 4.1.177
+date: 27:09:2024 07:15
 saved-by: Chris Bielow
 auto-generated-by: OBO-Edit 2.3.1
 default-namespace: MS
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -22523,6 +22523,20 @@ def: "Software designed to group multiple mass spectra by high similarity, gener
 is_a: MS:1000531 ! software
 
 [Term]
+id: MS:1003407
+name: Scout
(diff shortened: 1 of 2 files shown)


cbielow:
Do we need to increase the version number? It's just a re-ordering of terms. Content is identical.

edeutsch (changes requested):
yes, I think the GitHub actions/checking requires every change to have a bump in version number.

cbielow:
good to go

edeutsch (approved):

mobiusklein (approved):
Good catch, I had not considered the CV being unsorted by accession when I wrote the niche finder.

It is highly improbable that this breaks things for people directly consuming the OBO file.

PR #332, opened 2024-09-17
Updated psi-ms.obo to include the Kojak algorithm and its results.

mhoopmann:
I would like to suggest these additions to the psi-ms.obo to support results from Kojak. Please double-check that I've made all the appropriate relationships and syntax, as I used a text editor to make my changes...

### DIFF
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -1,7 +1,7 @@
 format-version: 1.2
-data-version: 4.1.175
-date: 25:08:2024 07:15
-saved-by: Joshua Klein
+data-version: 4.1.176
+date: 17:09:2024 15:26
+saved-by: Michael Hoopmann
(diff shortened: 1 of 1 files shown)


edeutsch (approved):

mobiusklein:
As-is the terms look good to me, but with the recent additions for mzIdentML 1.3 with crosslinking, I want to verify:

1. Are these scores all derived from single spectra, spectrum pairs, or either? In either of the latter cases, it would make sense to declare an `is-a` relation with MS:1003334.
2. Do any of these terms refer to one-or-the-other sequences in a crosslinked match, or is this an aggregate? If an aggregate, I would have to go read the paper in more detail  to understand what "consecutive ion matches" means. Otherwise, I can see how these link up with MS:1002511.

mobiusklein (approved):

PR #335, opened 2024-09-27
Add Exploris GC 240

edeutsch:


psi-ms.obo
```diff
@@ -22642,6 +22642,12 @@ is_a: MS:1001143 ! PSM-level search engine specific statistic
 relationship: has_order MS:1002108 ! higher score better
 relationship: has_value_type xsd:int ! The allowed value-type for this CV term
 
+[Term]
+id: MS:1003423
+name: Orbitrap Exploris GC 240
+def: Orbitrap Exploris GC 240 Mass Spectrometer [PSI:MS]
```
vagisha, line 22648: Looks like this has to be a quoted string.  Sorry for omitting that in my request for this term. 
def: "Orbitrap Exploris GC 240 Mass Spectrometer." [PSI:MS]

vagisha (approved):
Thanks, Eric.

mobiusklein (approved):

PR #336, opened 2024-09-27
do we want to add "neutral"?

edeutsch:
This has been proposed by the reviewer of the mzSpecLib manuscript. Do we want to make this change? What implications are there for downstream tools and files?

### DIFF
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -1,6 +1,6 @@
 format-version: 1.2
-data-version: 4.1.175
-date: 25:08:2024 07:15
+data-version: 4.1.180
+date: 04:10:2024 07:15
 saved-by: Joshua Klein
 auto-generated-by: OBO-Edit 2.3.1
 default-namespace: MS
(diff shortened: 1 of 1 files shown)


psi-ms.obo
henryhlam, line 1: I would support this change if it doesn't break anything. The definition is clear enough but people may not bother to look at the definition. Better to be clear since the adduct ion mass (the "M+H" mass) is sometimes also called the "theoretical mass" in the context of MS.

chambm:
If we add it as a synonym instead of just changing the name, it makes the meaning clear without breaking old software that uses the "theoretical mass" name. But just renaming isn't that big a deal either; it's less impactful than obsoleting a term and replacing it with a new one.

bittremieux (approved):
@chambm So you would swap the changed name and the synonym around compared to the current proposal? Either seem fine to me, and as Henry said, not a bad thing for avoiding confusion.

javizca:
I agree with using a synonym. As mentioned, the important thing is not to break existing software.

chambm:
Yes, I think swapping the name is fine and put the old name as a synonym.

PR #337, opened 2024-09-27
Add new spectrum origin type

edeutsch:
As pointed out by reviewer of mzSpecLib, existing spectrum origin types are not adequate to classify this type of spectrum, commonly used for DIA analysis.

### DIFF
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -22548,6 +22548,12 @@ name: timsTOF Ultra 2
 def: "Bruker Daltonics timsTOF Ultra 2." [PSI:MS]
 is_a: MS:1003123 ! Bruker Daltonics timsTOF series
 
+[Term]
+id: MS:1003424
+name: theoretical m/z observed intensity limited peak spectrum
+def: "Spectrum for which the peaks are limited to a subset of known product ions that are important for subsequent identification, the m/z values are corrected to theoretical values, and the intensity values are experimentally derived." [PSI:PI]
+is_a: MS:1003072 ! spectrum origin type
+
 
 [Term]
 id: MS:4000000


psi-ms.obo
henryhlam, line 1: what about "reduced spectrum for peptide-centric scoring" or "transition list spectrum" or "selected known product ion spectrum"?

The definition is OK, except that it looks a bit run-on. May need a connector "whose" after the first comma.

bittremieux: I agree that the definition is ok, the name takes a bit more effort to parse on its own though.

Maybe instead something like "selected fragment observed intensity spectrum"? The "selected" indicates that only a limited set of peaks are included, the "fragment" implicitly indicates that it's known fragments, thus theoretical m/z, and the "observed intensity" indicates that the intensity values are determined experimentally.

mobiusklein: +1 for `"selected fragment observed intensity spectrum"`. It's succinct and conveys the concept and doesn't tie it to any particular domain.

javizca: why not making it a bit more self explanatory? Unless we really want a short name: "Spectrum containing selected fragment ion peaks and experimental intensity"

edeutsch: 1) "theoretical m/z observed intensity selected fragment spectrum"
2) "selected fragment theoretical m/z observed intensity spectrum"  +++

def: "Spectrum for which the peaks are limited to a subset of known product ions that are important for subsequent identification, whose m/z values are corrected to theoretical values, and intensity values are experimentally derived." [PSI:PI]

mobiusklein (approved):

PR #360, opened 2025-01-13
Replace quartile/quarter -> quantile

bittremieux:
Some QC terms that were defined as "quartiles" or "quarters" have now been replaced by more general "quantile" terms.

Fixes [mzQC#259](https://github.com/HUPO-PSI/mzQC/issues/259).

### DIFF
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -1,7 +1,7 @@
 format-version: 1.2
-data-version: 4.1.186
-date: 10:11:2024 08:30
-saved-by: Yasset Perez-Riverol
+data-version: 4.1.187
+date: 13:01:2025 17:58
+saved-by: Wout Bittremieux
 default-namespace: MS
 namespace-id-rule: * MS:$sequence(7,0,9999999)$
 namespace-id-rule: * PEFF:$sequence(7,0,9999999)$
@@ -22898,7 +22898,7 @@ relationship: has_value_type xsd:float ! The allowed value-type for this CV term
 id: MS:4000026
 name: fragment ppm deviation median
 def: "The median of the distribution of observed fragment mass accuracies (MS:4000072) [in ppm] of identified MS2 spectra after user-defined acceptance criteria (FDR) are applied." [PSI:MS]
-comment:  Factors like temperature, airflow, or instrument calibration can cause mass shifts.
+comment: Factors like temperature, airflow, or instrument calibration can cause mass shifts.
 is_a: MS:4000003 ! single value
 relationship: has_metric_category MS:4000008 ! ID based metric
 relationship: has_metric_category MS:4000022 ! MS2 metric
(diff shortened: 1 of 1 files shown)


julianu (approved):

psi-ms.obo
```diff
@@ -22939,7 +22939,7 @@ name: area under TIC in MS1
 def: "The area under the total ion current chromatogram (MS:1000235) of all MS1 spectra." [PSI:MS]
 comment: The metric informs about the dynamic range of the acquisition. Differences between samples of an experiment may indicate differences in the dynamic range and/or in the sample content. The signal can be affected by contamination, retention time shifts, or loss of hydrophilic/hydrophobic peptides.
 is_a: MS:4000003 ! single value
-relationship: has_metric_category MS:4000009 ! ID free
+relationship: has_metric_category MS:4000009 ! ID free metric
```
mwalzer, line 22942: :+1: good catch (for all ID free **metric**s)

psi-ms.obo
```diff
@@ -22998,8 +22998,8 @@ relationship: has_units UO:0000010 ! second
 
 [Term]
 id: MS:4000052
-name: XIC-Height quartile ratios
-def: "The log ratio of successive XIC height quartiles. The metric's value triplet represents the log ratios of XIC-height-Q2 to XIC-height-Q1, XIC-height-Q3 to XIC-height-Q2, XIC-height max to XIC-height-Q3." [PSI:MS]
+name: obsolete XIC-Height quartile ratios
```
mwalzer, line 23001: Maybe not change the name by adding obsolete? The other obsolete term in the _.obo_  didn't go through a name change either as far as I can find.

psi-ms.obo
```diff
@@ -23010,6 +23010,8 @@ relationship: has_metric_category MS:4000018 ! XIC metric
 relationship: has_value_type xsd:float ! The allowed value-type for this CV term
 relationship: has_value_concept STATO:0000105 ! log signal intensity ratio
 relationship: has_units UO:0000191 ! fraction
+is_obsolete: true
+replaced_by: MS:4000182
```
mwalzer, line 23014: :+1: exemplary

mwalzer: the `replaced_by` in all old terms I mean

mwalzer (changes requested):
All good, except the name change?

psi-ms.obo
mobiusklein: Leaving the names as-is is another layer of backwards compatibility, but it's not guaranteed the way the `id` is guaranteed to never change.

bittremieux: I got these instructions from the Protege website for obsoleting terms, but happy to modify it according to conventions adopted by the PSI-MS CV.

edeutsch (approved):

PR #391, opened 2025-06-20
Add QC metrics

### LINKED ISSUE #387: CV terms for QC outlier detection

The following terms relate to different methods for flagging low-quality experiments using outlier detection on QC metrics:

```
[Term]
id: MS:4000XXX
name: Run outlier score PCA distance
def: "Outlier score for MS runs based on the median Euclidean distance for each run to every other run after robust PCA transformation of the QC metrics for all runs." [doi:10.1021/ac4034455, PSI:MS]
MS:4000005 ! table
relationship: has_metric_category MS:4000013 ! multiple runs based metric
relationship: has_column MS:4000086 ! mzQC input reference
relationship: has_column OBI:0200114 ! euclidean distance calculation

[Term]
id: MS:4000XXX
name: Run outlier score LoOP
def: "Outlier score for MS runs based on the local outlier probabilities (LoOP) algorithm applied to QC metrics from all runs." [doi:10.1021/acs.jproteome.6b00028, PSI:MS]
MS:4000005 ! table
relationship: has_metric_category MS:4000013 ! multiple runs based metric
relationship: has_column MS:4000086 ! mzQC input reference
relationship: has_column NCIT:C54154 ! Probability
```

Note that the OBI and NCIT ontologies (used for the outlier score columns) are currently not imported by the CV. But this is not necessary I think?.

bittremieux:
Fixes #387 and #388.

### DIFF
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -1,7 +1,7 @@
 format-version: 1.2
-data-version: 4.1.195
-date: 11:06:2025 21:31
-saved-by: David Cox
+data-version: 4.1.198
+date: 25:07:2025 21:40
+saved-by: Wout Bittremieux
 default-namespace: MS
 namespace-id-rule: * MS:$sequence(7,0,9999999)$
 namespace-id-rule: * PEFF:$sequence(7,0,9999999)$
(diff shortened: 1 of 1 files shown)


bittremieux:
Yes, these metrics are complete.

psi-ms.obo
```diff
@@ -24760,6 +24760,156 @@ relationship: has_units UO:0000191 ! fraction
...
+
+[Term]
+id: MS:4000194
+name: DIA isolation window count
+def: "The number of distinct DIA isolation windows (i.e. sharing the same low m/z boundary, high m/z boundary, and ion mobility parameters)." [PSI:MS]
+is_a: MS:4000003 ! single value
+relationship: has_metric_category MS:4000009 ! ID free metric
+relationship: has_metric_category MS:4000012 ! single run based metric
+relationship: has_metric_category MS:4000022 ! MS2 metric
+relationship: has_metric_category MS:1000792 ! isolation window attribute
```
mobiusklein, line 24827: This is distinct from `is_a: MS:1000792` so it is not legal to use in an actual isolation window? This is purely curiosity.

bittremieux: It's intended as a QC metric, so the relationship is just meant to add some more semantics. I didn't think about a use case for having it as an attribute for an actual isolation window as well. But I'm not sure it makes that much sense, because this is aggregated information across all isolation windows, rather than a property of a single isolation window.

edeutsch (approved):

PR #402, opened 2025-09-05
add MS instruments from MetaboLights, MetabolomicsWorkbench, GNPS and Instrument Vendor contacts

jonathan-hunter:
Hi all,

Please consider these additional MS instruments comprised of new unique instruments taken from MetaboLights, Metabolomics Workbench, GNPS and communications with instrument vendor representatives.

Best wishes,

Jonathan Hunter
Scientific Database Curator, Metabolomics, EMBL-EBI

### DIFF
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -23635,6 +23635,12 @@ def: "Shimadzu GCMS-TQ 8050 gas chromatograph - triple quadrupole mass spectrome
 is_a: MS:1000603 ! Shimadzu Scientific Instruments instrument model
 
 [Term]
+id: MS:1003576
+name: all ions fragmentation DIA alignment
+def: "Associating of product ions with the precursor from which they are derived, by comparing retention time and, where appropriate, ion mobility peaks." [PSI:MS]
+is_a: MS:1000543 ! data processing action
+
+[Term]
 id: MS:4000000
 name: PSI-MS CV Quality Control Vocabulary
 def: "PSI Quality Control controlled vocabulary term." [PSI:MS]
@@ -25774,7 +25780,7 @@ relationship: has_relation MS:4000103 ! number of identified quantification data
 
 [Term]
 id: MS:4000209
-name: peptide spectrum matches charges fractions 
+name: peptide spectrum matches charges fractions
 def: "The fraction of filtered peptide spectrum matches (PSMs) within the run for each specified charge state. The fractions [0,1] are given in the 'fraction' column, corresponding charges in the 'charge state' column. The highest charge state is to be interpreted as that charge state or higher. The numbers here are recorded after any filtering for PSM level quality (e.g. FDR filtering)." [PSI:MS]
 is_a: MS:4000005 ! table
 relationship: has_metric_category MS:4000009 ! ID based metric


ypriverol:
@copilot, can you double-check the following: 

- The new terms added in the instruments are not already in the specification. 
- If a new term for an instrument was added, and another model exists then they are not naming the same instrument differently. 
- That a model exist for each specific instrument?

edeutsch (commented):
There may be some duplicates here. We will review the list to see if some duplicates should be pruned (e.g. maXis HD and imact HD, are these the same thing?)

edeutsch:
I looked into the Velos and Elite situation, and here's my read of the situation:

Based on Thermo literature 
https://tools.thermofisher.com/content/sfs/manuals/VelosPro-Reference.pdf
, I see the following formal names for instruments:
LTQ Velos
LTQ Velos/ETD
Velos Pro
Velos Pro/ETD
LTQ Orbitrap Velos
Orbitrap Velos Pro
Orbitrap Elite

And so for the proposal and existing terms, here's what I think:

Proposed in #402:
name: Velos Pro			good to add
name: LTQ Orbitrap Velos ETD	good to add and maybe insert "/"
name: Orbitrap Velos Pro	Do not add. Update existing term to match this name (remove the "LTQ" portion which seems inconsistent with Thermo literature)
name: Orbitrap Elite		Do not add. Update existing term to match this name (remove the "LTQ" portion which seems inconsistent with Thermo literature)

Already in the CV:
name: LTQ Velos			good
name: LTQ Velos ETD		good or insert "/"
name: LTQ Orbitrap Velos	good
name: Velos Plus		?? This is synonym of the Orbitrap Velos Pro ??
name: LTQ Orbitrap Velos Pro	          Remove the "LTQ" part to be consistent with Thermo literature
name: LTQ Orbitrap Elite	            Remove the "LTQ" part to be consistent with Thermo literature

What do you think?

jonathan-hunter:
Thanks @ypriverol @chambm @edeutsch. 

- Removed duplicate entry for "Agilent 6546 Q-TOF LC/MS quadrupole time-of-flight mass spectrometer."
- Removed "Agilent CE-TOFMS capillary electrophoresis - time-of-flight mass spectrometer." Does not relate specifically to an MS instrument.
- Agreed RE: Velos Plus, couldn't find any reference to this existing based on a bit of googling. _How should we handle this one?_ Happy to implement the change.
- Added new unique Shimadzu instruments after communication with Vendor rep.
- Updated to "LTQ Orbitrap Velos/ETD".
- Removed "Orbitrap Velos Pro"
- Removed "Orbitrap Elite"
- Removed "Bruker micrOTOF-Q IITM QTOF", doesn't appear to be a real instrument.
- Re-indexed MSID on new terms.

Pre-existing:
- Updated MS:1000856 to "LTQ Velos/ETD". 
- Updated MS:1003096 to "Orbitrap Velos Pro".
- Updated MS:1001910 to "Orbitrap Elite".

edeutsch (approved):
This all seems good to me, thank you for working through all this. version still needs to be corrected.

edeutsch:
I think it should be 4.1.201

It would still be great to get an approval from @chambm @mobiusklein and anyone else willing to do a final check.

chambm:
I see a lot of definitions with double spaces (search for "  "). I'm looking at the file in Protege, and I just noticed "Q Exactive GC Orbitrap" is the only model that's a direct child of "Thermo Fisher Scientific instrument model"; other modern instruments are under "Thermo Scientific instrument model". Is IonSpec supposed to be a child of Varian? It's a bit odd we have a "Shimadzu MALDI-TOF instrument model" but also have some "MALDI 80xx" instruments in the sibling category "Shimadzu Scientific Instruments instrument model".

A bit of cleanup for Bruker too:
image
tims* should go to timsTOF model and impact II VIP should go to micrOTOF.

APEX-Qe 9.4T should be stylized like the other apex entries: "apex-Qe" and I'm not sure it should include the magnet strength. The Qe can apparently be coupled with 3 different magnet strengths, so if we're going to put the strength there we should do all 3, right? Actually I can't figure out the difference between "apex-Q" and "apex-Qe" - are they the same thing? Or they can use the same magnet strengths?

jonathan-hunter:
Yes, looks like Varian was indeed purchased by Agilent [1, 2].

1. https://www.reuters.com/article/innovationNews/idUSTRE56Q1ZC20090727/
2. https://en.wikipedia.org/wiki/Varian,_Inc.

edeutsch:
Ready to merge.
Note the changes that will affect ProteoWizard @chambm 
and ThermoRawFileParser @caetera

Issue #414, opened 2025-10-17
InstaNovo software


BioGeek:
### What is the new software's name?

InstaNovo

### Describe the software briefly.

InstaNovo is a deep learning based tool using a Transformer architecture for _de novo_ peptide sequencing. https://github.com/instadeepai/instanovo

### Software type

MS:1001456 ! analysis software

### Describe any additional information.

Publication: InstaNovo enables diffusion-powered de novo peptide sequencing in large-scale proteomics experiments https://doi.org/10.1038/s42256-025-01019-5

PR #416, opened 2025-10-17
Add InstaNovo and InstaNovo+ software

BioGeek:
Closes https://github.com/HUPO-PSI/psi-ms-CV/issues/414 (instaNovo) and https://github.com/HUPO-PSI/psi-ms-CV/issues/415 (InstaNovo+).


mobiusklein (approved):
Looks good to me.

PR #424, opened 2025-10-31
Update psi-ms.obo

henryhlam:
Distinguish monoisotopic and average mass for "adduct ion mass" and "theoretical neutral mass"

Also add unit to "adduct ion mass"


mobiusklein:
The new monoisotopic terms are just for completeness? They are otherwise redundant with the existing "generic" mass terms otherwise. Or is the concern that if the term is not as precise as possible in its definition it will permit ambiguity?

edeutsch (commented):
Looks good. approved at call today. need to update term ids and version information

mobiusklein (approved):

edeutsch (approved):

Issue #439, opened 2025-11-21
Reinstate acquisition level positive/negative ion mode terms


jonathan-hunter:
Dear All,

I'm writing to discuss the possibility of reinstating the positive and negative ion mode terms (see below). As part of the MetabolomicsHub project, we need terms for per-inject/per-acquisition polarity to capture positive/negative/alternating [LC-]MS acquisitions.

With that in mind, I would propose to re-instate the below terms, define them as per-inject/per-acquisition polarity modes and add `is_a: MS:1000857 ! run attribute` to the positive/negative terms as per `alternating polarity mode`.

Best wishes,

Jonathan
EMBL-EBI

```
[Term]
id: MS:1000076
name: negative ion mode
def: "OBSOLETE." [PSI:MS]
comment: This term was made obsolete because it was replaced by negative scan (MS:1000129).
is_obsolete: true

[Term]
id: MS:1000077
name: positive ion mode
def: "OBSOLETE." [PSI:MS]
comment: This term was made obsolete because it was replaced by positive scan (MS:1000130).
is_obsolete: true

[Term]
id: MS:1002833
name: alternating polarity mode
def: "Polarities of the scans of a run are alternating, i.e. both positive and negative mode scans are acquired." [PSI:PI]
is_a: MS:1000857 ! run attribute
```

jonathan-hunter:
> I think this is a fine idea as long as we make clear that the old terms are run level attributes and not scan level attributes. It's useful at a run level to know whether all the scans will be positive, negative, or mixed.

Thank you for the feedback. I will assemble this into a PR next week and we can fine-tune the implementation.

PR #442, opened 2025-11-24
Reinstate pos, neg polarity mode

jonathan-hunter:
Dear All, 

Please review the follow PR to reinstate positive and negative polarity modes, at the per run level, to the CV as previously discussed in #439 and the PSI-MS call.

These have been modified from the original obsolete forms to match `MS:1002833 ! alternating polarity mode` and all three terms move under the parentage `is_a: MS:1000524 ! data file content` as suggested in the call.

Best wishes,

Jon
EMBL-EBI

### DIFF
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -1,5 +1,5 @@
 format-version: 1.2
-data-version: 4.1.218
+data-version: 4.1.219
 date: 01:12:2025 13:00
 saved-by: Jonathan Hunter
 default-namespace: MS


chambm:
@edeutsch Should we rename the terms something more descriptive than "mode" - like "scanning mode" or something like that? Also, should we be reusing these obsolete terms, or should they be new terms?

jonathan-hunter:
@chambm @edeutsch I feel like we might be putting too much emphasis on the 'scan' element, and misrepresenting other important elements such as the polarity of the source ionisation process, ion optics and transmission, mass analyser configuration - all of which are affected by the instrument polarity mode.

I'm also thinking that keeping the current term names, is in line with the "alternating polarity mode" term, which is long-standing in the CV and would be better unchanged.

chambm:
That's a fair point about polarity affecting more than just the scanning (mass analyzer) part! What about "acquisition mode" or just "acquisition"? Is "alternating polarity mode" already used by one of our standards? I don't think it's used in mzML.

chambm:
I'm happy with acquisition (no mode). Question: can there me a "mixed polarity acquisition" where it's not really alternating, but not purely one or the other? Like the first half positive, second half negative, as a trivial example.

jonathan-hunter:
> I'm happy with acquisition (no mode). Question: can there me a "mixed polarity acquisition" where it's not really alternating, but not purely one or the other? Like the first half positive, second half negative, as a trivial example.

Yeah this is absolutely possible, and common in targeted analyses where you would have MRMs and associated ionisation settings in time windows. A good catch. Not so common in Metabolomics, but very much so in like bioanalysis, forensic tox., regulatory etc.

jonathan-hunter:
- Updated to 'xxx polarity acquisition' term names.
- Added 'mixed polarity acquisition'.

jonathan-hunter:
Personally, i like 'mixed polarity' and 'alternating' separate. I think this kind of rapid-switching of polarity is a discrete, well used and marketed functionality of newer mass spectrometers and warrants its own term.

Issue #377, opened 2025-04-13
Round 2 for ZSTD compression


mobiusklein:
### Describe the new term or terms you would like to add.

Revisits #167 and #168

Following on the discussion from the PSI Spring Workshop, since mzMLb did not receive uptake from the community for all the previously discussed reasons, we will revisit adding ZSTD compression to mzML.

The conversation proposes:
1. Provide ZSTD compression with byte shuffling with a "trade name". 
2. Provide delta encoding followed by ZSTD compression with byte shuffling with a "trade name".
3. Do not propose separate ZSTD without byte shuffling because it introduces an extra complexity.

I ran some experiments, generate points from 100.0 to 2100.0 for double precision floats spaced by 0.01 using ZSTD compression level 9 and zlib compression level 9: 

|   | Raw| Zlib| Shuffle + Zstd | Delta + Shuffle + Zstd |
|--|--------|--------|--------|--------|
|  Bytes  | 1600000 | 570704 | 12040 |  560 | 

I did not run a timing experiment because that will be language-sensitive.

### Begin Bikeshedding

We wanted to use a trade name because it seemed to the room that a verbose name would be harder to understand.

Some randomly generated names:

1. ZSTD-Shuffle - The anti-trade name just call it what it is
2. MZSTD - The least creative adaptation of the domain
3. MZD - I was wrong, I could be less creative

mobiusklein:
Some brief notes on complexity:

The byte shuffling algorithm is fairly straight-forward if you are willing to allocate a second array. In principle it is possible to implement a ping-pong shuffling algorithm that doesn't allocate but it will be _very_ easy to corrupt memory with it.

The delta encoding algorithm is very easy to implement in-place or with allocation. I used the algorithm from mzMLb because the Numpress Linear algorithm does not really support byte shuffling (at least not the one I had access to).

mobiusklein:
@chambm This conversation happened at the spring workshop, prompted by Sander Willems (@sander-willems-bruker) during the MzML for Ion Mobility and DIA session.

I'm willing to build this into ProteoWizard if you think this is acceptable. I believe there's already a static ZSTD library in the ProteWizard build tree so it shouldn't be a big change in maintenance burden.

mobiusklein:
As is traditional, I took a random RAW file I had lying around on disk and ran some conversion benchmarks:

[...]

PR #444, opened 2025-11-30
feature: add zstd compression terms

mobiusklein:
This starts to knit together the loose threads from #377. These will be the accessions I use in the mzML for DIA and ion mobility update text.


edeutsch:
@chambm would you have a look and provide approval or post concerns?

edeutsch:
@chambm one final nudge to weigh in before we merge?

PR #447, opened 2025-12-02
Remove wrong synonym for MS1 median cycle time

### LINKED ISSUE #446: mzQC: wrong synonym in  MS1 median cycle time
Hi,

`MS:4000192 name: MS1 median cycle time` has a synonym: `"sampling frequency" RELATED [MS:1000029]`
in https://github.com/HUPO-PSI/psi-ms-CV/blob/8fdedad42c6b87d37da7a46eacea0e52d8e58504/psi-ms.obo#L26782

That is a wrong synonym, since the `sampling frequency` is_a detector attribute, and refers to the frequency at which the detector (e.g. an Analog-to-Digital converter (ADC) ) scans ions within one spectrum. 

The `MS1 median cycle time` is related to the spectrum per seconds. 

I suggest removing that synonym.

Yours,
Steffen

sneumann:
Removed synonym 'sampling frequency' which is wrong.  
Wout agrees and I forgot to create the PR earlier today. Closes #446

### DIFF
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -26782,7 +26782,6 @@ relationship: has_value_type xsd:float
 id: MS:4000192
 name: MS1 median cycle time
 def: "The median time between consecutive MS1 scans across the full run." [PSI:MS]
-synonym: "sampling frequency" RELATED [MS:1000029]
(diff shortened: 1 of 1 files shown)


mobiusklein:
This is superceded by #453

Issue #435, opened 2025-11-06
CV for Chromatographic Columns?


jonathan-hunter:
Dear All,

I wanted to reach out ahead of the call tomorrow to open discussion around the description of chromatographic column models.

We at EBI Metabolomics are working with partners at GNPS/MassIVE and Metabolomics Workbench to unify open data standards in the field as part of the MetabolomicsHub project.

Where appropriate, we plan to use the PSI-MS-CV as the harmonised ontology for dataset metadata and have contributed MS instrument and separation system terms to enable this. Thank you for supporting and reviewing the pull requests from this work!

Moving forwards, another important variable is the description of chromatographic column model/chemistry. We would like to adapt the column database [1] assembled by Michael Witting and group as part of their RepoRT [2] project into a controlled vocabulary. We have discussed this with him and he is in support. The size of the database is noteworthy at approx. 50K entries.

With this in mind, I'd like to invite any thoughts on how best to achieve this and explore whether PSI-MS-CV would or would not be an appropriate home for this. I'll be on the PSI MS call tomorrow to discuss further.

Best wishes,
Jonathan
EMBL-EBI

[1] https://github.com/michaelwitting/RepoRT/tree/master/resources/column_database
[2] https://www.nature.com/articles/s41592-023-02143-z

javizca:
This seems to me like a good option. I don't know anything about the structure of the CV. I saw that the OBO version is from 2009, so very old really.

The usual challenge with re-starting a CV is to be able to include it in platforms like OLS and BioPortal, but other than having to put some work in the structure, and the disadvantage of having to handle another CV, this option (sepCV) is probably the best one.

-> One thing I don't quite understand is: what are these 50k things? Is it really "chromatographic column model/chemistry"? Are there 50k relevant to that?

The way things are organised in the database is separating different instances of columns by a combination of the brand of the column and then other characteristics such as pore and particle size, inner diameter, the usp classification and also by the column name. There are currently 49110 instances (see https://github.com/michaelwitting/RepoRT/blob/master/resources/column_database/column_database.tsv).

jonathan-hunter:
Some example data here for convenience:

company | column | pore size | particle size | inner diameter | length | usp | mode | Column name | full description | Metabolights Term | comment
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
[...]

PR #449, opened 2025-12-05
Hierarchy for chromatographic columns

jonathan-hunter:
Hi All,

As initially raised in #435 and discussed in PSI-MS calls, I have begun to put together the parent terms and hierarchy to capture chromatographic columns in the CV.

For now, I have considered anything to do with chromatography except the chromatography system itself as a 'run attribute'. This has the drawback of having attributes of attributes, which feels a bit off. We may rather want to pull the chromatography columns/phases out to a higher level and consider them to be physical entities. This kind of depends on how you guys want to view chromatography in relation to MS (the primary focus of the CV).

Feedback welcome!

### DIFF
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -25355,6 +25355,7 @@ id: MS:1003924
 name: pore size
 def: "The average diameter of the pores within the particles of the stationary phase, typically represented in angstroms (Å)" [PSI:MS]
 is_a: MS:1003923 ! chromatographic stationary phase attribute
+relationship: has_units UO:0000019 ! angstrom
 relationship: has_value_type xsd:string ! The allowed value-type for this CV term
 
 [Term]


chambm:
USP codes and chromatography mode would likewise be attributes but could be defined explicitly because there aren't too many of them. Then each one could have a proper definition.

jonathan-hunter:
I think if additions of this length are too verbose then this is a good alternative.

Could argue that the RepoRT DB is that auxiliary file already and validate permutations based on that.

jonathan-hunter:
I have removed the example instance of a chromatographic column and its vendor parent for now. I also took steps to resolve merge conflicts.

Let's discuss approving this for merge in the call later, please!

edeutsch (approved):
Seems pretty good to me. Should "pore size" be a float with a unit?

It feels like "liquid chromatographic column" should be a child of "chromatographic column", but I see why you've organized it the way it is.. there is ample precedent.

PR #453, opened 2025-12-05
Fix merge-ability of #447

mobiusklein:


mobiusklein:
This appears to be mangled due to a mismatch between Copilot merges and the `gh` command line tool to handle resolving merge conflicts.

PR #460, opened 2025-12-15
Add new data processing terms for zero intensity trimming

mobiusklein:
These terms describe an existing filter found in ProteoWizard's MSConvert `--filter "zeroSamples RemoveExtra 1-"` and an equivalent behavior is used by mzPeak's zero intensity trimming and null marking for profile data.

Context:

Some vendors, including Thermo, at least some of the time will export spectra where there are long runs of zero intensity value data points. These vendors can internally represent these zero intensity intervals inexpensively, but in order to export them to mzML, they must be expanded in memory and written to disk. While the intensity portion is cheaply compressed away, the m/z portion is for all intents and purposes just as complicated from an information theoretic perspective as the rest of the spectrum.  IIRC, this feature was introduced for `mz5` back around 2012, code living [here](https://github.com/ProteoWizard/pwiz/blob/42d530e25b5c09076bd94218028ca208c620da83/pwiz/analysis/common/ZeroSampleFiller.cpp).

I was irked when this behavior also silently flipflopped on/off automatically in Waters vendor libraries a few years ago (https://github.com/ProteoWizard/pwiz/issues/1166). This would at least let us register that this occurred or should occur. Undecided on how to express how to fill back in those missing points as yet and any model parameters involved.

### DIFF
--- a/.github/workflows/validate-obo.yml
+++ b/.github/workflows/validate-obo.yml
@@ -27,6 +27,9 @@ jobs:
     # Runs a single command using the runners shell
     - name: Validate OBO file
       run: fastobo-validator --duplicates psi-ms.obo
+    - uses: actions/setup-python@v6
+      with:
+          python-version: '3.10'
     - name: Check sorting
       run: |
         python -m pip install fastobo


edeutsch:
So then maybe this to be completely clear?
```
[Term]
id: MS:1003901
name: zero intensity data point trimming
def: "Apply an algorithm to remove excess zero intensity value data points by retaining only zeros that flank non-zero intensity value data points from a profile spectrum." [PSI:MS]
is_a: MS:1000543 ! data processing action
```

mobiusklein:
Technically, the retention of flanking zeros may or may not occur as in the ProteoWizard filter flanking zeros are retained for profile data but not for centroid data. This is a "reasonable" decision because flanking zero points in centroid data are not meaningful under the common interpretation of centroids, while they are for profile data as they contribute estimation of peak width.

I'll update the descriptions to reflect this distinction soon, but I also should think about what this means w.r.t. ion mobility too. At least outside my implementation(s).

chambm:
I like Eric's description. I don't see the terms in the current change set though?

mobiusklein:
Thanks for catching that. It must have gone awry during the merge conflict resolution and then I had to switch tasks.

mobiusklein:
An approval will trigger the merge.

PR #465, opened 2026-01-15
Update sample introduction terms

jonathan-hunter:
Please consider the following proposed improvements to terms within 'inlet type'.

**Ontology term definition improvements:**
* Updated the definition of `MS:1000058` (`flow injection analysis`) to clarify the method and added the synonym "FIA".
* Updated the definition of `MS:1000060` (`infusion`) for precision and added the synonym "direct infusion".
* Revised the definition of `MS:1000249` (`direct liquid introduction`) to provide historical context and technical detail.

PS The 'inlet type' class has a bit of a mix of things including inlet _components_, sample introduction _methods_, ionization _types_ and might be worth revisiting at some point.

### DIFF
--- a/.github/workflows/validate-obo.yml
+++ b/.github/workflows/validate-obo.yml
@@ -27,3 +27,10 @@ jobs:
     # Runs a single command using the runners shell
     - name: Validate OBO file
       run: fastobo-validator --duplicates psi-ms.obo
+    - uses: actions/setup-python@v6
+      with:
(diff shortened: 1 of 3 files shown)


mobiusklein (approved):
These details look good to me. Unfortunately I encounter these rarely in practice because they rely on ProteoWizard being able to infer them.

edeutsch (approved):

PR #502, opened 2026-04-09
Add LECO PEG and JEOL JPF file format terms (MS:1003973-1003974)

jonathan-hunter:
Please review!

What do you guys think of the precedent of synonym EXACT for the file extension?

### DIFF
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -1,6 +1,6 @@
 format-version: 1.2
 data-version: 4.1.247
-date: 13:03:2026 12:00
+date: 17:04:2026 12:00
 saved-by: Jonathan Hunter
 default-namespace: MS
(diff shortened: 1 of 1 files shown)


chambm:
I don't think exact synonym is right. Consider .raw with Waters .raw and Thermo .raw. A "has_file_extension" and "has_directory_extension" relationship, perhaps? Probably not really necessary. It could just be part of the description.

jonathan-hunter:
> I don't think exact synonym is right. Consider .raw with Waters .raw and Thermo .raw. A "has_file_extension" and "has_directory_extension" relationship, perhaps? Probably not really necessary. It could just be part of the description.

Yeah I agree, I have revised the PR accordingly.

edeutsch (approved):

mobiusklein (approved):

Issue #501, opened 2026-04-08
`centroid peak`


oolonek:
We are looking for a way to represent each individual peak in a centroid spectrum. As far as we can tell we couldn't find the appropriate class in the MS ontology. A proposition could be:

`centroid peak`
is_a: [peak](http://purl.obolibrary.org/obo/MS_1000231)
relationship: part_of [centroid spectrum](http://purl.obolibrary.org/obo/MS_1000127)
def: "A zero-width peak defined by a pair of corresponding values from the [m/z array](http://purl.obolibrary.org/obo/MS_1000514) and [intensity array](http://purl.obolibrary.org/obo/MS_1000515) associated to the [centroid spectrum](http://purl.obolibrary.org/obo/MS_1000127)"

Happy to give additional informations or reuse existing concepts in case they already exist.

Proposed together with @llegregam and @Mattheex

jonathan-hunter:
It seems like a reasonable addition to me. Bearing in mind my response to #500 I would argue that an object cannot be part-of an attribute/quality, and therefore I would suggest you:
- Drop the relationship line.
- Consider revising the definition, for example:
`def: "A peak represented as a single discrete data point defined by a distinct m/z value and a corresponding intensity value.`

jonathan-hunter:
`centroid peak (MS:1003908)` merged in #508. Thank you for the contribution.

Issue #500, opened 2026-04-08
Update description of centroid spectrum


oolonek:
Currently, `centroid spectrum` (http://purl.obolibrary.org/obo/MS_1000127) description represents an action "Processing of profile data to produce spectra that contains discrete peaks of zero width. Often used to reduce the size of dataset." 
This is somehow confusing compared to the similar `profile spectrum` (http://purl.obolibrary.org/obo/MS_1000128) which describes an object "_A profile mass spectrum is created when data is recorded with ion current (counts per second) on one axis and mass/charge ratio on another axis._"

A more coherent definition could be "_A centroid spectrum resulting from the processing of profile data to produce spectra that contains discrete peaks of zero width. Often used to reduce the size of dataset._"

proposed together with @llegregam and @Mattheex

jonathan-hunter:
I agree that the current definition for `centroid spectrum` incorrectly reflects a process and there is room for improvement here.

`centroid spectrum` & `profile spectrum` are children of `spectrum representation` "Way in which the spectrum is represented, either with regularly spaced data points or with a list of centroided peaks." This is in turn part-of some `spectrum`.

My interpretation `centroid spectrum` [attribute/quality] is-a `spectrum representation` [attribute/quality] of `spectrum` [object]. i.e. `centroid spectrum` is an attribute/quality and not an object or process.

With this in mind, something like the following would be more accurate semantically:

`centroid spectrum`: "A spectrum representation in which each peak is recorded as a single discrete data point at a distinct m/z value with a corresponding intensity."
`profile spectrum`: "A spectrum representation in which ion signal is recorded as a continuous or regularly sampled distribution of intensity values across the m/z range."

I would imagine these terms are invoked by downstream applications like PWiz, mzML etc. so will need community input @chambm, @mobiusklein, @edeutsch - although this is a definition only change.

jonathan-hunter:
`centroid spectrum (MS:1000127)` & `profile spectrum (MS:1000128)` term definitions revised in #508. Thank you for the contributions.

PR #508, opened 2026-04-20
Refine PSI-MS spectrum definitions; add centroid peak; fix PEFF definition typos

jonathan-hunter:
PR to address suggestions in issues #500 & #501.

### DIFF
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -28527,7 +28527,7 @@ relationship: has_value_type xsd:string ! The allowed value-type for this CV ter
 [Term]
 id: PEFF:0001013
 name: ModRes
-def: "PEFF keyword for the modified residue without a PSI-MOD or UniMod identifier." [PSI:PEFF]
+def: "PEFF keyword for the modified residue without a PSI-MOD or Unimod identifier." [PSI:PEFF]
 is_a: PEFF:0000003 ! PEFF file sequence entry term
 relationship: has_regexp PEFF:1002001 ! regular expression for a value in a key-value pair of a PEFF description line describing one sequence position followed by one PEFF term name and one optional comment
 relationship: has_value_type xsd:string ! The allowed value-type for this CV term
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -28631,7 +28631,7 @@ relationship: has_value_type xsd:string ! The allowed value-type for this CV ter
 [Term]
 id: PEFF:0001027
 name: ModResUnimod
-def: "PEFF keyword for the modified residue with UniMod identifier." [PSI:PEFF]
+def: "PEFF keyword for the modified residue with Unimod identifier." [PSI:PEFF]
 is_a: PEFF:0000003 ! PEFF file sequence entry term
 relationship: has_regexp PEFF:1002001 ! regular expression for a value in a key-value pair of a PEFF description line describing one sequence position followed by one PEFF term name and one optional comment
 relationship: has_value_type xsd:string ! The allowed value-type for this CV term
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -1068,14 +1068,14 @@ is_a: MS:1000031 ! instrument model
 [Term]
 id: MS:1000127
 name: centroid spectrum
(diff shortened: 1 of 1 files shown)


jonathan-hunter:
Also includes a couple of bonus typo fixes in the PEFF term defs.

psi-ms.obo
```diff
@@ -1068,14 +1068,14 @@ is_a: MS:1000031 ! instrument model
...
-def: "Processing of profile data to produce spectra that contains discrete peaks of zero width. Often used to reduce the size of dataset." [PSI:MS]
+def: "A spectrum representation in which each peak is recorded as a single discrete data point at a distinct m/z value with a corresponding intensity." [PSI:MS]
 synonym: "Discrete Mass Spectrum" EXACT []
 is_a: MS:1000525 ! spectrum representation
 
 [Term]
 id: MS:1000128
 name: profile spectrum
-def: "A profile mass spectrum is created when data is recorded with ion current (counts per second) on one axis and mass/charge ratio on another axis." [PSI:MS]
+def: "A spectrum representation in which ion signal for each peak is recorded as a distribution of intensity values across the m/z range." [PSI:MS]
```
mobiusklein: Replace "m/z" with "coordinate".  We use `profile spectrum` to refer to UV spectra too.

```suggestion
def: "A spectrum representation in which ion signal for each peak is recorded as a distribution of intensity values across the coordinate range." [PSI:MS]
```

psi-ms.obo
```diff
@@ -1068,14 +1068,14 @@ is_a: MS:1000031 ! instrument model
 [Term]
 id: MS:1000127
 name: centroid spectrum
-def: "Processing of profile data to produce spectra that contains discrete peaks of zero width. Often used to reduce the size of dataset." [PSI:MS]
+def: "A spectrum representation in which each peak is recorded as a single discrete data point at a distinct m/z value with a corresponding intensity." [PSI:MS]
 synonym: "Discrete Mass Spectrum" EXACT []
```
mobiusklein, line 1072: We also have spectra which are centroided in the ion mobility dimension, so replacing "m/z" with "coordinate" would make sense here too.

psi-ms.obo
```diff
@@ -25667,6 +25667,12 @@ is_a: MS:1003214 ! mass spectrometry acquisition method aspect
 is_a: MS:1003213 ! mass spectrometry acquisition method
 relationship: part_of MS:1003905 ! targeted analysis
 
+[Term]
+id: MS:1003908
+name: centroid peak
+def: "A peak represented as a single discrete data point defined by a distinct m/z value and a corresponding intensity value." [PSI:MS]
```
mobiusklein:
```suggestion
def: "A peak represented as a single discrete data point defined by a distinct coordinate value and a corresponding intensity value." [PSI:MS]
```

psi-ms.obo
```diff
@@ -1068,14 +1068,14 @@ is_a: MS:1000031 ! instrument model
 [Term]
 id: MS:1000127
 name: centroid spectrum
-def: "Processing of profile data to produce spectra that contains discrete peaks of zero width. Often used to reduce the size of dataset." [PSI:MS]
+def: "A spectrum representation in which each peak is recorded as a single discrete data point at a distinct m/z value with a corresponding intensity." [PSI:MS]
```
mobiusklein:
```suggestion
def: "A spectrum representation in which each peak is recorded as a single discrete data point at a distinct coordinate value with a corresponding intensity." [PSI:MS]
```

mobiusklein (changes requested):
These are relatively minor.

psi-ms.obo
edeutsch: maybe since "coordinate" is a somewhat ambiguous term in this context, we could change "coordinate" to "coordinate (typically m/z or ion mobility)"? This provides the reader with a hint that coordinate typically means m/z but could be other things?

pabinz (commented):
thx for the clean-up of the PEFF defs

edeutsch (approved):
Looks great, thanks.

Maybe while we're in here fixing things, it would be nice to change UniMod to Unimod in line 28530. The M does not seem to be capitalized by them https://www.unimod.org/unimod_help.html

mobiusklein (approved):

PR #511, opened 2026-05-04
Added new term peak group

### LINKED ISSUE #510: Include 'peak group' within HUPO PSI CV

The peak group concept is pivotal to the analysis and quality control of SWATH proteomic experiments. (see https://pubmed.ncbi.nlm.nih.gov/28188537/.) It should be captured within the controlled vocabulartory as it serves as a container to organize signals during data analysis.

jannickk:
This pull request resolves #510

### DIFF
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -28296,6 +28296,13 @@ relationship: has_column MS:1003044 ! number of missed cleavages
 relationship: has_column UO:0000191 ! fraction
 
 [Term]
+id: MS:4000216
+name: peak group 
+def: "a peak group is any collection of extracted ion traces that exhibit the same peak profile as they derive from the same precursor ion. For instance, extracted ion traces of the fragment ions of a peptide form a peak group as do isotopologues from one metabolite precursor ion" [PSI:MS]
+relationship: has_value_type xsd:string ! The allowed value-type for this CV term
+
+
+[Term]
 id: PEFF:0000001
 name: PEFF CV term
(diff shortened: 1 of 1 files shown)


psi-ms.obo
```diff
@@ -28295,6 +28295,13 @@ relationship: has_metric_category MS:4000008 ! ID based metric
 relationship: has_column MS:1003044 ! number of missed cleavages
 relationship: has_column UO:0000191 ! fraction
 
+[Term]
+id: MS:4000216
```
jonathan-hunter, line 28299: MS:1003975

psi-ms.obo
```diff
@@ -28295,6 +28295,13 @@ relationship: has_metric_category MS:4000008 ! ID based metric
...
+name: peak group 
+def: "a peak group is any collection of extracted ion traces that exhibit the same peak profile as they derive from the same precursor ion. For instance, extracted ion traces of the fragment ions of a peptide form a peak group as do isotopologues from one metabolite precursor ion" [PSI:MS]
```
jonathan-hunter, line 28301: Suggest `as they derive from a common chemical origin` - isotopologues do not derive from the same precursor ion exactly.

Also I suggest you add references to in-source fragmentations and adducts/clusters - these would be important in a metabolomics context.

Could consider synonyms such as 'molecular feature group' or similar.

Could consider adding in that they are determined as having correlated XIC profiles.

jonathan-hunter (changes requested):

jonathan-hunter:
Hey @jannickk, I have merged #513, shall we progress this one? Thank you!

PR #513, opened 2026-05-04
Added retention time window target

### LINKED ISSUE #512: Add retention time window target to HUPO-PSI CV

While there are terms for retention time window lower offset and upper offset, a term for the target retention time is missing.

jannickk:
resolves #512

### DIFF
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -28296,6 +28296,15 @@ relationship: has_column MS:1003044 ! number of missed cleavages
 relationship: has_column UO:0000191 ! fraction
 
 [Term]
+id: MS:4000217
+name: retention time window target
+def: "The target of the retention time window." [PSI:MS]
+is_a: MS:1000915 ! retention time window attribute
+relationship: has_units UO:0000010 ! second
+relationship: has_units UO:0000031 ! minute
+relationship: has_value_type xsd:float ! The allowed value-type for this CV term
+
+[Term]
 id: PEFF:0000001
 name: PEFF CV term
 def: "PSI Extended FASTA Format controlled vocabulary term." [PSI:PEFF]
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -1,7 +1,7 @@
 format-version: 1.2
-data-version: 4.1.248
-date: 20:04:2026 12:01
-saved-by: Jonathan Hunter
+data-version: 4.1.249
+date: 22:05:2026 08:24
+saved-by: Jannick Kappelmann
 default-namespace: MS
 namespace-id-rule: * MS:$sequence(7,0,9999999)$
 namespace-id-rule: * PEFF:$sequence(7,0,9999999)$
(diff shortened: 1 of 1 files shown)


psi-ms.obo
```diff
@@ -28295,6 +28295,15 @@ relationship: has_metric_category MS:4000008 ! ID based metric
 relationship: has_column MS:1003044 ! number of missed cleavages
 relationship: has_column UO:0000191 ! fraction
 
+[Term]
+id: MS:4000217
+name: retention time window target
+def: "The target of the retention time window." [PSI:MS]
```
jonathan-hunter: def: "The reference retention time from which a retention time window is defined. The target may be derived from a predicted, expected, observed, or user-specified retention time." [PSI:MS]

jonathan-hunter (changes requested):

psi-ms.obo
jonathan-hunter: Edit suggested to avoid circular logic.

psi-ms.obo
```diff
@@ -28295,6 +28295,15 @@ relationship: has_metric_category MS:4000008 ! ID based metric
 relationship: has_column MS:1003044 ! number of missed cleavages
 relationship: has_column UO:0000191 ! fraction
 
+[Term]
+id: MS:4000217
```
jonathan-hunter: MS:1003976

jonathan-hunter (changes requested):
Should be moved into the main MS CV term id series, I believe.

jonathan-hunter:
Hi @jannickk, Thank you for the contribution.

I think this makes sense as an addition and I don't see any obvious collisions, indeed `MS:1000916` & `MS:1000917` already reference a "target retention time" in their definitions so it is nice to also formalise that anchor.

I also notice that `retention time` is-a `peptidoform attribute`, we may want to widen this out to include other analyte types in the future, but this isn't within the scope of this PR.

One could argue that it should be `target retention time` is-a `retention time` but I think it being considered a constituent of the retention time window is also fine and in-keeping with the existing pattern.

psi-ms.obo
jannickk: @jonathan-hunter I adapted the definition

psi-ms.obo
jannickk: I adapted the id, I was not aware that there a particular logic attached to the ids

jannickk:
May we close this pull request first? I will then rebase and proceed with PR https://github.com/HUPO-PSI/psi-ms-CV/pull/511

jonathan-hunter:
Hi @jannickk, thanks for making the revisions. The PSI-MS group will meet on Friday and discuss this PR then, but I think this looks good and would expect it to be merged shortly after the meeting.

mobiusklein (approved):
This looks consistent, thank you for the addition.

jonathan-hunter:
@edeutsch @mobiusklein Hi both, are we happy to merge this?

edeutsch (approved):

edeutsch:
Approved at the call. @jonathan-hunter please merge when convenient.

jonathan-hunter (approved):

PR #520, opened 2026-06-08
Added some new terms for use in mzTab

douweschulte:
These are some new terms for use in mzTab:
- Protein sequence: let me know if `MS:1001344|AA sequence` is not a good unit t use here, also if `has_value_type` is used correctly here.
- `amino acid`- and `token confidence level`: let me know if you have a better name, these should be fine but might not be  the easiest to gain the exact usage of based solely on the name, maybe something like `local PSM confidence level at the amino acid level` might be better but I did not manage to find something concise.

I though I would leave the version bump to somewhat later after the discussions are done.

### DIFF
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -26080,6 +26080,29 @@ relationship: has_units UO:0000031 ! minute
 relationship: has_value_type xsd:float ! The allowed value-type for this CV term
 
 [Term]
+id: MS:1003984
+name: amino acid confidence level
+def: "The confidence level of the PSM assignment per amino acid, where 0 is not confident and 1 is maximally confident." [PSI:MS]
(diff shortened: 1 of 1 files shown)


mobiusklein (commented):
Looks good to me, pending version bump. Will leave until Friday to allow for others to comment.

edeutsch (commented):
Looks generally good. "token" is very generic, making "token confidence level" very ambiguous. What about something like "peptidoform token confidence level" or something like that? i.e. can we make "token" more specific in the name?

douweschulte:
I like the name 'peptidoform token confidence level' better. Thanks for the suggestion. The version is bumped now as well and I merged it with the main branch and put it in the right location.

mobiusklein (approved):

edeutsch (approved):

Issue #524, opened 2026-06-10
Issue with escaped characters


douweschulte:
I have been playing with creating an Obo file formatter to guarantee consistent ordering  and formatting and found an issue with the escaping in regular expressions. Some expressions are properly escaped but many are not. Some examples are below:
```
[Term]
id: MS:1001957
name: (?<=[ALIV])(?!P)
def: "Regular expression for leukocyte elastase." [PSI:PI]
is_a: MS:1001180 ! Cleavage agent regular expression
```
Any Obo parser SHOULD parse this with the `!P)` being a comment. OLS parses this as such: https://www.ebi.ac.uk/ols4/ontologies/ms/classes/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FMS_1001957.
```
[Term]
id: MS:1002231
name: regular expressions for a GUID
def: "([A-Fa-f0-9]\{8\}-([A-Fa-f0-9]\{4\}-)\{3\}[A-Fa-f0-9]\{12\})." [PSI:PI]
is_a: MS:1002479 ! regular expression
```
OLS takes the `\` characters as escape characters (as defined in the OBO format) and renders this as such: `([A-Fa-f0-9]{8}-([A-Fa-f0-9]{4}-){3}[A-Fa-f0-9]{12})` https://www.ebi.ac.uk/ols4/ontologies/ms/classes/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FMS_1002231. 
```
[Term]
id: MS:1002480
name: regular expression for a digital object identifier (DOI)
def: "(10[.][0-9]\{4,\}(?:[.][0-9]+)*/(?:(?![\"&\'<>])[^ \t\\r\n\\v\\f])+)." [PSI:PI, http://dx.doi.org/]
is_a: MS:1002479 ! regular expression
```
OLS interprets this as `(10[.][0-9]{4,}(?:[.][0-9]+)*/(?:(?!["&'<>])[^ 	\r
\v\f])+).` (hard to see but it contains a literal new line not the regex escaped version) https://www.ebi.ac.uk/ols4/ontologies/ms/classes/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FMS_1002480.

douweschulte:
Other terms with similar escaping issues: `MS:1003206`, `MS:1003825` and nearly everything with a regular expression.

[...]

PR #534, opened 2026-06-23
Escape backslashes

douweschulte:
Related: #524

This PR fixes some unescaped backslashes. The entire file is valid and preserves the intended backslashes using obo-robo after this is merged. There are two classes of cases that will change, but those where 'overescaped' which is allowed by the spec but not necessary:

```
def: "Definition." [Wikipedia:http\://en.wikipedia.org/wiki/Electric_field]
xref: binary-data-type:MS\:1000521 "32-bit float"
```

Which will transform into:
```
def: "Definition." [Wikipedia:http://en.wikipedia.org/wiki/Electric_field]
xref: binary-data-type:MS:1000521 "32-bit float"
```
I can likely change this to have these automatically escaped as well if there is a need for that (not needed by the spec though I think).

Slightly unrelated news: I now have a `keep-ordering` (still reorders line types but preserves everything else) option in obo-robo to make it easier to integrate the formatting workflow piecemeal.

### DIFF
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -1,6 +1,6 @@
 format-version: 1.2
-data-version: 4.1.256
-date: 11:06:2026 17:15
+data-version: 4.1.257
+date: 23:06:2026 22:00
 saved-by: Douwe Schulte
 default-namespace: MS
 namespace-id-rule: * MS:$sequence(7,0,9999999)$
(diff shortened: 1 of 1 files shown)


mobiusklein (approved):
Verified that all the escapes are necessary for proper parsing using `fastobo`

edeutsch (approved):

PR #536, opened 2026-07-02
feature: add unsigned integer binary array data types, index array type

mobiusklein:
This PR may be controversial as it adds a few new binary array data types.

The main motivation here is to add an `index array` type that works as a parallel array (alongside `m/z array`, for example) that connects points in an array to another entity. This is the use-case for mzPeak, at least. This array holds an unsigned integer of some width, so I've added terms for those as well.

This is where things become problematic, as those new binary data types pose a storage problem for existing implementations if someone uses them in mzML somehow. Several implementations use a per-type list of arrays, one for each category of data type, which gets you static typing and storage, but if there's no compatible storage, there's nowhere to put them. They _could_ be stored as signed integers, but we would need a secondary channel to denote that these values are unsigned.

Many instruments write integer valued intensities, and integers compress more easily, so I added them as valid value types for that as well.

If this is too big an issue, I can go back and re-design this component to be based entirely on Arrow data types so that it doesn't depend upon the CV, but it would be easier to inter-convert with mzML.

### DIFF
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -1,5 +1,5 @@
 format-version: 1.2
-data-version: 4.1.257
+data-version: 4.1.258
 date: 23:06:2026 22:00
 saved-by: Douwe Schulte
 default-namespace: MS
@@ -3808,6 +3808,10 @@ name: intensity array
 def: "A data array of intensity values." [PSI:MS]
 xref: binary-data-type:MS\:1000521 "32-bit float"
 xref: binary-data-type:MS\:1000523 "64-bit float"
+xref: binary-data-type:MS\:1003850 "64-bit unsigned integer"
+xref: binary-data-type:MS\:1003851 "32-bit unsigned integer"
+xref: binary-data-type:MS\:1000519 "32-bit integer"
+xref: binary-data-type:MS\:1000522 "64-bit integer"
 is_a: MS:1000513 ! binary data array
 relationship: has_units MS:1000131 ! number of detector counts
 relationship: has_units MS:1000132 ! percent of base peak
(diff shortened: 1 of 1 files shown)


edeutsch (commented):
okay as far as I'm concerned, but I don't know the implications of this for downstream code. @chambm should definitely weight in.

endianness does not apply to 8-bit integer? endian only relevant for multi-byte integers?

chambm:
I'm not really sure it's worth the complexity to make implementors support unsigned as well as signed integers. Also, I think the decision was made years ago to stick with float for intensity arrays because the formats that would especially benefit from integer compression weren't using the array data type (e.g. MsNumpress). Does MzPeak really need both signed and unsigned integer arrays? I mean, a signed 64-bit index array will take you to 9,223,372,036,854,775,807. I don't think we're going to be having arrays that big in MzPeak any time soon, I hope?

mobiusklein:
RE 8-bit endian, yes, that's copy-pasta. I didn't have a data type for storing bitflags, which is how one of Thermo's peak dimensions is encoded. The ASCII term wouldn't work since it specifically uses null-terminated strings.

>I'm not really sure it's worth the complexity to make implementors support unsigned as well as signed integers.  ... Does MzPeak really need both signed and unsigned integer arrays? I mean, a signed 64-bit index array will take you to 9,223,372,036,854,775,807. I don't think we're going to be having arrays that big in MzPeak any time soon, I hope?

No, I don't think we're going to get there. I don't even think we're going to top out signed 32-bit integers any time soon unless we go back to writing our IM-MS frames/cycles unrolled again. As it is, Java and C# don't even support more than signed 32-bit indices without using FFI. This was for 1:1 correctness with the Parquet/Arrow storage. It's probably OK to use the signed 32/64-bit integer term and add a note to the spec about signed-ness.

I can also just not try to use PSI-MS terms for this purpose, and leave it up to the implementation to decide how to pass data to an API that expects one of the ~3 (float64, float32, int32) types of arrays mzML has conditioned us to expect.

>Also, I think the decision was made years ago to stick with float for intensity arrays because the formats that would especially benefit from integer compression weren't using the array data type (e.g. MsNumpress)

This one is useful outside of mzPeak, but I am requesting it for semantic validity/correctness _for_ mzPeak.

It amounts to a 2-5% improvement with plain `zlib` compression for profile spectra when you compare a lossless compression of intensity arrays. Naturally, just truncating the fractional portion of the number makes it a lot more compressible too. 

Take ye olde `small.mzML` and pretend it comes from an instrument that would be compatible by discarding the fractional part of the intensity values, and turning them back into floats for demonstration.
```python
import zlib
import numpy as np
from pyteomics import mzml

spec = next(mzml.read("small.mzML"))
print(
    # first convert to int to discard the fractional part, then convert back to float32 for the byte layout
    len(zlib.compress(spec['intensity array'].astype(int).astype(np.float32), 9)) /  
    len(zlib.compress(spec['intensity array'].astype(np.int32), 9))
)
# 1.0428724174800814

print(
    len(zlib.compress(spec['intensity array'].astype(int).astype(np.float32), 9)) / 
    # combine with byte shuffling, pardon the abuse of NumPy arrays
    len(zlib.compress((spec['intensity array'].astype(np.int32)).view((np.uint8)).reshape((-1, 4)).T.flatten(), 9))
)
# 1.3325106870565422
```
You see comparable results using Zstandard compression. As it is today, you can already use [byte-shuffled zstd compression](http://purl.obolibrary.org/obo/MS_1003781) so the only code changes would be the integer conversion. Of course, MS-Numpress is still better for minimizing the loss while achieving similar compression. If the data were already integers to begin with though, there is no loss.

chambm:
Supporting integer intensity arrays in pwiz would probably require significant refactoring. Maybe something to consider after I finish the pwiz-sharp port.

edeutsch (approved):

mobiusklein:
@chambm I removed the integer intensity component as well. Do you have any other comments? Otherwise I'll merge this Monday.

PR #530, opened 2026-06-18
add chromatographic column models from RepoRT

jonathan-hunter:
Hi All, 

As previously discussed in #435, please consider this PR to add Liquid chromatography columns from the RepoRT database.
@michaelwitting has kindly given permission to re-use the database for this purpose.

As previously discussed, this is a significant addition, comprising a roughly 50% increase in the total size of the CV by terms/lines.

To my knowledge, there is no other ontology for chromatographic columns (or systems) and this would therefore make PSI-MS-CV the de-facto ontology for the description of MS coupled chromatographic methods alongside MS instruments. I have put these in the hierarchy as a 'run attribute' and tried to reflect the pattern used elsewhere in the CV.

Also included are two new typdefs `has_separation_mode` & `usp_designation` and these are used in the column model terms. `<manufacturer> chromatographic column model` terms have been added within a new `chromatographic column model` parent.

Within a new MS:5000000 namespace, the column model terms are placed (everything else is in the normal MS:1000000 series).

This PR is a draft at this time, I need to do QC and plan to try and populate the missing usp/retention modes, but I wanted to re-open the discussion at the earliest opportunity. You may recall that the RepoRT database also includes permutations of column dimensions resulting in a combinatorial explosion up to 50,000 terms - this information has not been included as previously discussed.

Best wishes,
Jon

PR #538, opened 2026-07-28
Chrom. Columns PR1/2 - Split off psi-ms-core.obo, introduce psi-ms-columns.obo-fragment and merge wiring.

jonathan-hunter:
PR 1 of 2 for chrom. columns update.
This PR is intended to be merged first with the following objectives:

- psi-ms.obo becomes a read-only build output.
- psi-ms-core.obo becomes the community created, human-edited controlled vocabulary object. 
- psi-ms-columns.obo-fragment sets the initial state of the chromatographic columns import from Repo-RT. 
- build-release-obo.yml and build_release_obo.py splices psi-ms-columns.obo-fragment into psi-ms-core.obo to build psi-ms.obo.
- Other minor edits have been made to release and validation flows to support this design.
- A temporary manual PR gate has been added to require maintainer review before the merged psi-ms.obo is committed.

I have validated psi-ms-columns.obo-fragment against repo-RT and expect it to be stable.
I have structured this with a view to easy diff comparison of output artefacts (especially psi-ms.obo) in github, so we can be sure that the core CV is 100% preserved.

PR of 2 of 2 will follow and will contain the remainder of draft PR #530 - mainly the psi-ms-columns.obo-fragment generator and RepoRT polling logic.

Please review with a view to merging this first step soon!


jonathan-hunter:
I've tested the flows in my fork and they are working as intended. Note the last commit fixes a pre-existing bug in upstream/master where the GitHub release contains OBO file version N alongside OWL file version N-1.

edeutsch (approved):

jonathan-hunter:
**TODO: Merge latest psi-ms.obo contents into PR ps-ms-core.obo prior to PR merge.** If we miss this, recent terms will be dropped from the release (although this will be visible in the manual review gate before commiting psi-ms.obo [combined] and making a release).

scripts/build_release_obo.py
```diff
@@ -0,0 +1,69 @@
...
+
+import argparse
+import re
+import sys
+
+CORE_DEFAULT = "psi-ms-core.obo"
+FRAGMENT_DEFAULT = "psi-ms-columns.obo-fragment"
+OUTPUT_DEFAULT = "psi-ms.obo"
+
+MS_ID_LINE = re.compile(r"^id: MS:\S+$", re.M)
```
mobiusklein: Why is the multi-line modifier used here? This looks like it should all be on one line.

scripts/build_release_obo.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+def check_fragment(fragment_text, fragment_path):
```
mobiusklein, line 20: This looks like rough validation of OBO syntax, why not use `fastobo`? It can parse a file without a header.

.github/workflows/make-release-on-tag.yml
```diff
@@ -1,27 +1,32 @@
...
+        run: cp psi-ms.owl ms.owl
+
       - name: Create a GitHub release
         uses: ncipollo/release-action@v1
         with:
-          tag: v${{ env.CV_VERSION }}
-          name: Release v${{ env.CV_VERSION }}
+          tag: ${{ github.ref_name }}
+          name: Release ${{ github.ref_name }}
           artifacts: "psi-ms.obo,*ms.owl"
```
mobiusklein, line 30: We should still publish the core OBO as a subset without the column data.

.github/workflows/update-owl.yaml
```diff
@@ -49,20 +50,68 @@ jobs:
...
         with:
           token: ${{ secrets.PSI_MS_CV_GITHUB_TOKEN }}
           pull-request-number: ${{ steps.cpr.outputs.pull-request-number }}
           merge-method: squash
+
+  # Tags the OWL merge commit, which is the first commit holding both the OBO and the OWL
+  # built from it. Tagging in generate-owl tagged the psi-ms.obo commit one earlier, and
+  # make-release-on-tag checks out the tag, so every release shipped the previous OWL.
+  tag:
+    if: ${{ contains(github.event.head_commit.message, 'Update OWL file') }}
```
mobiusklein, line 70: Good catch

scripts/build_release_obo.py
jonathan-hunter: `MS_ID_LINE.finditer(core_text)` returns an interator comprised of all matches to `^id: MS:\S+$` across all lines in `core_text`. Each match is a single line `^....$` but `core_text` is the whole block so without  re.MULTILINE* it will only match `^` on the start of the first line for example - returning no matches.

The for loop then passes over all lines matching the regex and variable `last` is read by the code below to find the end of the [Term] stanza.

*I'll change re.M to re.MULTILINE for readability.

scripts/build_release_obo.py
jonathan-hunter: The `.github/workflows/build-release-obo.yml` flow calls `fastobo.load` on the `psi-ms.obo` after the splice, which would raise `SyntaxError` on an invalid OBO.

This one is just an earlier check to catch a header in the fragment that would end up mid-file. I'll add "Full OBO syntax validation is performed downstream by fastobo." to the docstring.

.github/workflows/make-release-on-tag.yml
jonathan-hunter: New `psi-ms-core.obo & psi-ms-core.owl` release files? Or add subsetdef for 'core' and let downstream parsers handle it? Or both?

.github/workflows/update-owl.yaml
jonathan-hunter: Thanks for the review Joshua, hope to discuss with you online shortly.

jonathan-hunter:
@mobiusklein @edeutsch If we want to use OBO subset tagging of terms, we will need to add `subset: core` to every applicable term. This will increase the size of the `psi-ms-core.obo` by approx 10-20% addition lines. We could should probably also add a `subsetdef:` for the mzQC set also. Sound good?

edeutsch:
oh, hm. I was not expecting this. So is it true that if we use *any* subset properties, then *all* terms must have subset properties? That seems unfortunate

It would be okay. But I am leaning toward just removing all subset terms then. 

Is it useful? My first reaction was not useful. But now that I ponder a bit more, we really do have a lot of subsets now.. NCIT, core, mzQC, PEFF, columns. Maybe subsetting all would actually help out software that only cared about some terms? maybe it would be handy be able to write software like:
```
cv.load(subsets=['core', 'PEFF'])
```
Other opinions?

jonathan-hunter:
Sorry, I wasn't clear in my explanation, I didn't intend to imply that all terms must have a subset defined to use the subset function. Just that in order for the `core` subset to function and contain the `core` terms, then those `core` terms must contain `subset: core`. There doesn't seem to be an alternative way to bulk assign them.

jonathan-hunter:
Tested the workflow in my fork and reviewed the output artefacts (psi-ms-core.obo and psi-ms.obo).

I think this is good to go. Questions for approval in today's call - are we happy with the `subset:` and `subsetdef:` setup here?

If so, I hope to merge this in a breakout with @mobiusklein and baby sit it through the process. If we have another PR approved and ready to merge after, this would be good to observe also - we will need to move changes from psi-ms.obo to psi-ms-core.obo before commit, for pre-existing PRs.

jonathan-hunter:
Ready to merge. 

Reminder - the workflow will raise a PR for the combined `psi-ms.obo` output artefact, so we can review it before committing and triggering a release.

PR #541, opened 2026-08-04
Reparent "calibration spectrum" to "spectrum attribute"

### LINKED ISSUE #539: Calibration spectrum hides the type of spectrum

In ProteoWizard we're looking at labeling Waters lockmass spectra as calibration spectra now, but discovered that because calibration spectrum is a spectrum type, it cannot coexist with the MS1 spectrum term (according to semantic mapping). We're thinking that that term should probably be a spectrum attribute instead of part of the spectrum type hierarchy.

mobiusklein:
Closes #539

### DIFF
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -1,7 +1,7 @@
 format-version: 1.2
-data-version: 4.1.258
+data-version: 4.1.259
 date: 23:06:2026 22:00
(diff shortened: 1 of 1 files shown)


edeutsch (approved):

edeutsch (approved):

chambm (approved):
Thanks!

PR #544, opened 2026-08-13
Chrom. Columns PR2/2 - Add the repo-rt column generator, its checks and the weekly sync

jonathan-hunter:
**Depends on #538 — merge that first and babysit the first obo merge and release.** Merging in order also means every future generator output is diffed against a validated initial state.

## What this adds
- `scripts/chrom_columns/generate_psi_ms_columns.py` — builds `psi-ms-columns.obo-fragment` from the [RepoRT column database](https://github.com/michaelwitting/RepoRT/blob/master/resources/column_database/column_database.tsv). Currently 65 vendors, 1863 models, 1928 terms, in the **MS:5000000–MS:5999999** band.
- `.github/repo-rt/psi-ms-column-ids.tsv` — the id ledger, committed here at its initial state. Keyed on `(company, column)` and append-only: a retired id is never reused, and a model returning to the catalog reclaims its original id.
  This is the two-way sync contract with RepoRT, which mirrors assigned ids back into a `psi_ms_id` catalog column — but **MS:ID authority stays here**.
- `.github/workflows/chrom_columns_sync_repo-rt.yml` — the weekly sync (Mondays 06:00 UTC, plus manual `workflow_dispatch`).
- `scripts/check_aggregate.py`, `scripts/chrom_columns/check_column_mapping.py`,`scripts/chrom_columns/check_no_duplicate_ids.py` — validation the sync job runs before it will open a PR.

## How the sync behaves
Each run regenerates the fragment and the ledger, validates them (fastobo parse, sort order, no duplicate ids, ledger/fragment consistency, and a trial splice into`psi-ms.obo` checked for unresolved references), patch-bumps `data-version` and refreshes `date` in `psi-ms-core.obo` when the fragment changed, and opens a PR.

**Nothing is committed to master automatically** — the PR must be reviewed, approved and merged by a human. When the regenerated files are identical to the committed ones, no PR is opened at all.

The PR body carries the full id-ledger change list (minted / renamed / retired / resurrected) plus any upstream data-quality problems to fix in RepoRT.

## Downstream
Once a sync PR is merged, the release chain follows automatically: `psi-ms.obo` is rebuilt, then `psi-ms.owl`, the tag, and the published release. For the initial roll-out #538 keeps a temporary human gate on the rebuilt `psi-ms.obo` as well; that gate is intended to be removed once the flow has proven itself.

This PR supersedes #530 and is _intended_ to be the "final" version, barring revisions.

### DIFF
--- a/scripts/chrom_columns/check_column_mapping.py
+++ b/scripts/chrom_columns/check_column_mapping.py
@@ -1,0 +1,105 @@
+#!/usr/bin/env python3
+"""Validate .github/repo-rt/psi-ms-column-ids.tsv (the id ledger) against the fragment.
+
+The ledger is the sole authority for id assignment -- generate_psi_ms_columns.py reads
+it before minting -- and the cross-repo contract repo-rt joins on to back-propagate
+ids. Rows are append-only: a model that leaves the catalog keeps its row and its id is
+never reused, so the fragment's terms are a SUBSET of the ledger, not equal to it.
+Vendor terms have rows too, with an empty `column` field; vendor and retired rows can
+never match a catalog row (every catalog row has a non-empty column), so they are
+inert to repo-rt's join.
+
+This checks: an exact 3-column header; every data row has three tab-separated fields;
+vendor rows (empty column) carry in-band vendor ids and model rows in-band leaf ids;
+ids and (company, column) keys are unique; and every term id in the fragment has a
+ledger row -- an unledgered term would mean an id was minted outside the ledger.
+
+Usage:
+    python scripts/chrom_columns/check_column_mapping.py <mapping.tsv> <module.obo>
+"""
+import re
+import sys
+from collections import Counter
+
+HEADER = "company\tcolumn\tpsi_ms_id"
+# Must match generate_psi_ms_columns.py's VENDOR_BAND / LEAF_BAND and LEDGER_HEADER.
+# Kept as literals so this checker stays stdlib-only (the generator needs pandas);
+# if the bands or header ever change, update both files together.
+VENDOR_LO, VENDOR_HI = 5000000, 5000999
+LEAF_LO, LEAF_HI = 5001000, 5999999
+ID_RE = re.compile(r"^MS:(\d{7})$")
+
+
+def ids_in_obo(obo_path):
+    """(vendor_ids, leaf_ids, errors) from every `id: MS:` line of the fragment.
+    A malformed id line is reported through errors rather than crashing the run."""
+    vendors, leaves, errors = set(), set(), []
+    for n, line in enumerate(open(obo_path, encoding="utf-8"), start=1):
+        if not line.startswith("id: MS:"):
+            continue
+        m = ID_RE.match(line[len("id: "):].strip())
+        if not m:
+            errors.append(f"{obo_path} line {n}: malformed id line {line.strip()!r}")
+        elif VENDOR_LO <= int(m.group(1)) <= VENDOR_HI:
+            vendors.add(m.group(0))
+        elif LEAF_LO <= int(m.group(1)) <= LEAF_HI:
+            leaves.add(m.group(0))
+        else:
+            errors.append(f"{obo_path} line {n}: id {m.group(0)} outside the vendor/leaf bands")
+    return vendors, leaves, errors
+
+
+def validate(mapping_path, obo_path):
+    errors = []
+    lines = open(mapping_path, encoding="utf-8").read().split("\n")
+    if lines[0] != HEADER:
+        errors.append(f"header must be {HEADER!r}, got {lines[0]!r}")
+    keys, vendor_ids, leaf_ids = [], [], []
+    for n, line in enumerate(lines[1:], start=2):
+        if line == "":
+            continue
+        fields = line.split("\t")
+        if len(fields) != 3:
+            errors.append(f"line {n}: expected 3 tab-separated fields, got {len(fields)}")
+            continue
+        company, column, ms_id = fields
+        keys.append((company, column))
+        lo, hi, kind, bucket = ((VENDOR_LO, VENDOR_HI, "vendor", vendor_ids) if column == ""
+                                else (LEAF_LO, LEAF_HI, "leaf", leaf_ids))
+        m = ID_RE.match(ms_id)
+        if not m or not (lo <= int(m.group(1)) <= hi):
+            errors.append(f"line {n}: psi_ms_id {ms_id!r} is not an in-band {kind} id")
+        else:
+            bucket.append(ms_id)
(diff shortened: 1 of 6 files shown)


scripts/chrom_columns/generate_psi_ms_columns.py
```diff
@@ -0,0 +1,1035 @@
...
+import os
+import re
+import sys
+from collections import Counter, defaultdict
+from pathlib import Path
+
+import pandas as pd
+
+OUTPUT_DEFAULT = "psi-ms-columns.obo-fragment"
+INPUT_DEFAULT = ".jeh-local/column_database_fixed.tsv"
```
mobiusklein, line 77: Is this a local file path, or is it populated by one of the other scripts?

scripts/chrom_columns/check_no_duplicate_ids.py
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env python3
+"""Fail if an OBO file contains duplicate term ids.
```
mobiusklein, line 2: This looks redundant with `fastobo-validator`  with `--all` or `--duplicates` flags, does it work around something?

scripts/chrom_columns/check_column_mapping.py
```diff
@@ -0,0 +1,105 @@
...
+        elif LEAF_LO <= int(m.group(1)) <= LEAF_HI:
+            leaves.add(m.group(0))
+        else:
+            errors.append(f"{obo_path} line {n}: id {m.group(0)} outside the vendor/leaf bands")
+    return vendors, leaves, errors
+
+
+def validate(mapping_path, obo_path):
+    errors = []
+    lines = open(mapping_path, encoding="utf-8").read().split("\n")
```
mobiusklein, line 54: Why not use `csv`? The standard library's CSV parser is fully compliant and frees you of needing to worry about future escape characters, quotes, and line-endings.

scripts/chrom_columns/check_column_mapping.py
```diff
@@ -0,0 +1,105 @@
...
+VENDOR_LO, VENDOR_HI = 5000000, 5000999
+LEAF_LO, LEAF_HI = 5001000, 5999999
+ID_RE = re.compile(r"^MS:(\d{7})$")
+
+
+def ids_in_obo(obo_path):
+    """(vendor_ids, leaf_ids, errors) from every `id: MS:` line of the fragment.
+    A malformed id line is reported through errors rather than crashing the run."""
+    vendors, leaves, errors = set(), set(), []
+    for n, line in enumerate(open(obo_path, encoding="utf-8"), start=1):
```
mobiusklein, line 37: `fastobo` would alleviate a lot of text manipulation here.

scripts/chrom_columns/generate_psi_ms_columns.py
```diff
@@ -0,0 +1,1035 @@
...
+        "\ufeff": "",  # zero-width no-break space / BOM
+    }
+)
+
+
+def clean(text):
+    """Normalize a catalog cell for use as an OBO name/def value and as an identity key:
+    drop control characters, fold invisible/non-breaking characters, collapse whitespace."""
+    text = "".join(ch for ch in text if ch >= " " and ch != "\x7f")
+    return re.sub(r"\s+", " ", text.translate(INVISIBLE)).strip()
```
mobiusklein, line 177: Does this need to be done using only those characters and not the full range of Unicode whitespace? Why not use `unicodedata.normalize("NFKD", text)` from the standard library?

scripts/chrom_columns/generate_psi_ms_columns.py
```diff
@@ -0,0 +1,1035 @@
...
+    Because QUOTE_NONE does no quote processing, a field an upstream export wrapped in
+    double quotes (the CSV convention for a value containing a comma) keeps those quotes
+    as literal characters. In the identity columns they would leak into term names and
+    — since a quoted name differs from the previously assigned unquoted one — mint fresh
+    ids and churn the mapping; in the value columns they corrupt the emitted annotation
+    instead (a `usp` cell of '"L1, L11"' yields the codes '"L1' and 'L11"'). Either way
+    the file is not the tab-delimited catalog this reads, so a wrapped-quote cell in any
+    required column ABORTS the run."""
+    try:
+        text = Path(tsv_path).read_bytes().decode("utf-8")
```
mobiusklein, line 204: Is there a reason this doesn't just use the `csv` module? The `csv.DictReader` type would handle all of these cases and would be easily converted to a `pd.DataFrame` without re-parsing the text.

scripts/chrom_columns/generate_psi_ms_columns.py
```diff
@@ -0,0 +1,1035 @@
...
+    """End a sentence with a period unless it already does (vendors ending 'Inc.')."""
+    return sentence if sentence.endswith(".") else sentence + "."
+
+
+def escape_def(text):
+    """Escape a value for an OBO quoted def: string (upstream names may contain " or \\)."""
+    return text.replace("\\", "\\\\").replace('"', '\\"')
+
+
+def escape_tag(text):
```
mobiusklein, line 445: All of these escapes could be handled with `fastobo`, at the added cost of having to use the ten distinct types instead of text manipulation.

scripts/chrom_columns/generate_psi_ms_columns.py
```diff
@@ -0,0 +1,1035 @@
...
+    its old id. Identity lives here, not in the fragment: names there are display labels
+    (escaped, collision-suffixed) and are never read back.
+
+    ponytail: a mirror-proven rename CONSUMES the old key's row (apply_mirror rebinds
+    it), so a rename reverted upstream mints a second id for the same physical column.
+    Retiring the old key instead needs a fourth ledger field, which is the cross-repo
+    contract repo-rt's add_psi_ms_id.py joins on -- add it there and here together, on
+    evidence that renames actually get reverted.
+    """
+    if not os.path.exists(path):
```
mobiusklein, line 592: What is the name lifecycle this alludes to? Why wouldn't tracking synonyms be sufficient?

jonathan-hunter:
@mobiusklein Many thanks for the reviews here and on #538. I'm on leave this Friday, so I'll miss the PSI-MS call this week. I'll aim to get these revisions/responses done by mid next week so we can discuss next Friday - if you are available.

jonathan-hunter:
> @mobiusklein Many thanks for the reviews here and on #538. I'm on leave this Friday, so I'll miss the PSI-MS call this week. I'll aim to get these revisions/responses done by mid next week so we can discuss next Friday - if you are available.

Replied to the reviews on #538, I have some other pressing work ahead of the call later so will return to this one next week.

PR #545, opened 2026-08-23
add 11 ID free QC metrics for metabolomics

fatemehmirzade:
add 11 new ID free QC metrics in the MS:4000216-4000226 range inserted after MS:4000215.

discussed and approved in [HUPO-PSI/mzQC#328](https://github.com/HUPO-PSI/mzQC/issues/328) by @cbielow, @julianu and @bittremieux. header updated (data-version, date, saved-by)per the README.

### DIFF
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -1,7 +1,7 @@
 format-version: 1.2
-data-version: 4.1.258
-date: 23:06:2026 22:00
-saved-by: Douwe Schulte
+data-version: 4.1.259
+date: 23:08:2026 12:00
+saved-by: Fatemeh Mirzadeh Sarcheshmeh
 default-namespace: MS
 namespace-id-rule: * MS:$sequence(7,0,9999999)$
 namespace-id-rule: * PEFF:$sequence(7,0,9999999)$
@@ -28,6 +28,7 @@ remark: creator: David Cox <david.cox <-at-> sciex.com>
 remark: creator: Jonathan Hunter <jhunter <-at-> ebi.ac.uk>
 remark: creator: Samuel Wein <sam <-at-> openms.de>
 remark: creator: Jeroen Van Goey <j.vangoey <-at-> instadeep.com>
+remark: creator: Fatemeh Mirzadeh Sarcheshmeh <fatemeh.mirzadehsarcheshmeh <-at-> uantwerpen.be>
 remark: creator: Jannick Kappelmann <jannick.kappelmann <-at-> gmail.com>
 remark: publisher: HUPO Proteomics Standards Initiative Mass Spectrometry Standards Working Group and HUPO Proteomics Standards Initiative Proteomics Informatics Working Group
 remark: When appropriate the definition and synonyms of a term are reported exactly as in the chapter 12 of IUPAC orange book. See http://www.iupac.org/projects/2003/2003-056-2-500.html and http://mass-spec.lsu.edu/msterms/index.php/Main_Page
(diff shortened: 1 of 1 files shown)


edeutsch (approved):

psi-ms.obo
```diff
@@ -28556,6 +28557,142 @@ relationship: has_metric_category MS:4000008 ! ID based metric
...
+relationship: has_metric_category MS:4000009 ! ID free metric
+relationship: has_metric_category MS:4000012 ! single run based metric
+relationship: has_metric_category MS:4000022 ! MS2 metric
+relationship: has_value_type xsd:float ! The allowed value-type for this CV term
+relationship: has_units MS:1000040 ! m/z
+
+
+[Term]
+id: MS:4000224
+name: MS2 precursor isolation window width
```
jonathan-hunter: MS2 precursor isolation window **median** width?

psi-ms.obo
```diff
@@ -28556,6 +28557,142 @@ relationship: has_metric_category MS:4000008 ! ID based metric
...
+relationship: has_metric_category MS:4000012 ! single run based metric
+relationship: has_metric_category MS:4000021 ! MS1 metric
+relationship: has_value_type xsd:float ! The allowed value-type for this CV term
+relationship: has_units MS:1000040 ! m/z
+relationship: has_value_concept STATO:0000035 ! range
+
+
+[Term]
+id: MS:4000223
+name: MS2 fragment acquisition range above precursor median
```
jonathan-hunter: suggest "MS2 median acquisition range above precursor" - currently implies median of precursor.

psi-ms.obo
```diff
@@ -28556,6 +28557,142 @@ relationship: has_metric_category MS:4000008 ! ID based metric
...
+relationship: has_metric_category MS:4000022 ! MS2 metric
+relationship: has_metric_category MS:1000792 ! isolation window attribute
+relationship: has_value_type xsd:float ! The allowed value-type for this CV term
+relationship: has_units MS:1000040 ! m/z
+
+
+[Term]
+id: MS:4000225
+name: number of distinct precursor m/z values
+def: "The number of distinct selected ion m/z values (MS:1000744) recorded in MS2 scan headers in a single MS run, where two values are considered identical if they agree after rounding to 3 decimal places." [PSI:MS]
```
jonathan-hunter: Should this be based on ppm rather than m/z?

psi-ms.obo
```diff
@@ -28556,6 +28557,142 @@ relationship: has_metric_category MS:4000008 ! ID based metric
...
+relationship: has_metric_category MS:4000012 ! single run based metric
+relationship: has_metric_category MS:4000022 ! MS2 metric
+relationship: has_value_type xsd:int ! The allowed value-type for this CV term
+relationship: has_units UO:0000189 ! count unit
+
+
+[Term]
+id: MS:4000226
+name: number of distinct isolation target m/z values
+def: "The number of distinct isolation window target m/z values (MS:1000827) programmed for fragmentation in a single MS run, where two values are considered identical if they agree after rounding to 3 decimal places." [PSI:MS]
```
jonathan-hunter: As above?

jonathan-hunter (commented):
Some thoughts...

psi-ms.obo
jonathan-hunter: By which I mean: the 3 d.p. quantisation is far more permissive about merging values in relative terms at low m/z and far stricter at high m/z (roughly 14 ppm at m/z 70 against 1 ppm at m/z 1000) — whereas mass accuracy is roughly constant in ppm across the range.

bittremieux: You're right, although a side note is that this is only applicable to Orbitrap instruments. TOFs have a consistent mass accuracy across the m/z range.

In this case we have these values for the pan-ReDU data, which consists of a wide diversity of instruments. So it's never going to be a perfect definition across all data. For simplicity we went with straightforward decimal rounding.

In fact, the values that we have are rounded at 2 decimals rather than 3, so this needs to be slightly updated @fatemehmirzade. It's a pretty high ppm-error for low-m/z measurements, but that's the data that we have at the moment.

psi-ms.obo
mobiusklein: Also explicitly state whether width is distance from the target or distance from lower bound to upper bound. 

There are some sad cases where instrument software errors led the isolation target m/z to fall outside the isolation window bounds themselves.

psi-ms.obo
```diff
@@ -28556,6 +28557,142 @@ relationship: has_metric_category MS:4000008 ! ID based metric
...
+relationship: has_metric_category MS:4000012 ! single run based metric
+relationship: has_metric_category MS:4000019 ! MS metric
+relationship: has_value_type xsd:int ! The allowed value-type for this CV term
+relationship: has_units UO:0000189 ! count unit
+
+
+[Term]
+id: MS:4000222
+name: MS1 observed m/z bounds
+def: "The minimum and maximum m/z among detected peaks in MS1 spectra in a single MS run, reported as a [min, max] tuple." [PSI:MS]
```
mobiusklein, line 28643: Similar to other terms in this PR, does it make sense to directly reference [`MS:1000528|lowest observed m/z`](http://purl.obolibrary.org/obo/MS_1000528) and it's sibling for highest observed m/z in the description? Or are those not trusted?

bittremieux: Yes, those relationships can definitely be added.

fatemehmirzade:
Thank you for the review and everything is addressed in 8408060b6a13eaecb002e60379809037a914327f now
1. MS:4000223 renamed to MS2 median acquisition range above precursor
2. MS:4000224 renamed to MS2 precursor isolation window median width also the definition now is stating that width is measured bound to bound (the sum of MS:1000828 and MS:1000829), independent of the target m/z
3. rounding changed from 3 to 2 decimal places in MS:4000225 and MS:4000226 matching the resolution of the our data
4. has_relation added to MS:4000222, MS:4000224, MS:4000225 and MS:4000226
5. header also bumped to 4.1.260 on top of the current master

jonathan-hunter (approved):

mobiusklein:
I have update the PR to address merge conflicts and bump the version number. I'll merge in a day if there are no other changes you want to make.

mobiusklein (approved):

PR #550, opened 2026-09-07
add mzQC metrics for ms imaging purposes

patrick-bo:
add 11 new metrics specific to MS Imaging into the MS:4000216 - MS:4000226 range.

This has been previously discussed in the mzqc working group (https://github.com/HUPO-PSI/mzQC/issues/336).

Header updated as per readme, IDs might need to be updated as https://github.com/HUPO-PSI/psi-ms-CV/pull/545 is currently aiming for the same namespace.

### DIFF
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -1,6 +1,6 @@
 format-version: 1.2
 data-version: 4.1.261
-date: 07:09:2026 7:30
+date: 07:09:2026 07:30
 saved-by: Patrick Boschmann
 default-namespace: MS
 namespace-id-rule: * MS:$sequence(7,0,9999999)$
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -1,7 +1,7 @@
 format-version: 1.2
-data-version: 4.1.260
-date: 01:09:2026 11:55
-saved-by: Jonathan Hunter
+data-version: 4.1.261
+date: 07:09:2026 7:30
+saved-by: Patrick Boschmann
 default-namespace: MS
 namespace-id-rule: * MS:$sequence(7,0,9999999)$
 namespace-id-rule: * PEFF:$sequence(7,0,9999999)$
@@ -29,6 +29,7 @@ remark: creator: Jonathan Hunter <jhunter <-at-> ebi.ac.uk>
 remark: creator: Samuel Wein <sam <-at-> openms.de>
 remark: creator: Jeroen Van Goey <j.vangoey <-at-> instadeep.com>
 remark: creator: Jannick Kappelmann <jannick.kappelmann <-at-> gmail.com>
+remark: creator: Patrick Boschmann <patrick.boschmann <-at-> uni-tuebingen.de>
 remark: publisher: HUPO Proteomics Standards Initiative Mass Spectrometry Standards Working Group and HUPO Proteomics Standards Initiative Proteomics Informatics Working Group
 remark: When appropriate the definition and synonyms of a term are reported exactly as in the chapter 12 of IUPAC orange book. See http://www.iupac.org/projects/2003/2003-056-2-500.html and http://mass-spec.lsu.edu/msterms/index.php/Main_Page
 remark: For any queries contact psidev-ms-vocab@lists.sourceforge.net
@@ -9278,6 +9279,7 @@ is_a: MS:1001249 ! search input details
 [Term]
 id: MS:1001412
 name: search tolerance plus value
+def: "Search tolerance above the theoretical value, given in relative or absolute units." [PSI:PI]
 is_a: MS:1001411 ! search tolerance specification
 relationship: has_units UO:0000166 ! parts per notation unit
 relationship: has_units UO:0000169 ! parts per million
@@ -9288,6 +9290,7 @@ relationship: has_value_type xsd:double ! The allowed value-type for this CV ter
 [Term]
 id: MS:1001413
 name: search tolerance minus value
+def: "Search tolerance below the theoretical value, given in relative or absolute units." [PSI:PI]
 is_a: MS:1001411 ! search tolerance specification
 relationship: has_units UO:0000166 ! parts per notation unit
 relationship: has_units UO:0000169 ! parts per million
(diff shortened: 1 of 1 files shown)


edeutsch (approved):

psi-ms.obo
```diff
@@ -28570,6 +28573,139 @@ relationship: has_metric_category MS:4000008 ! ID based metric
...
+relationship: has_metric_category MS:4000012 ! single run based metric
+relationship: has_metric_category MS:4000216 ! imaging metric
+relationship: has_units UO:0000191 ! fraction
+relationship: has_value_type xsd:float ! The allowed value-type for this CV term
+relationship: has_relation MS:1000505 ! base peak intensity
+
+[Term]
+id: MS:4000220
+name: base peak fraction image quantiles
+def: "The first to n-th quantile of the base peak fraction values over all acquired spectra of an imaging acquisition. A value triplet represents the quartiles Q1, Q2, Q3. The number of values in the tuple implies the quantile mode." [PSI:MS]
```
mobiusklein: "quantile mode" is vague here. Could you please be more explicit?

psi-ms.obo
```diff
@@ -28570,6 +28573,139 @@ relationship: has_metric_category MS:4000008 ! ID based metric
...
+comment: A compact summary of the base peak fraction image (MS:4000219). Not acquired positions and spectra with a total ion current of 0 are excluded from the calculation. Values are not comparable between centroid and profile spectra.
```
mobiusklein: The note about centroid vs. profile probably needs to be part of the definition, or at the very least made clearer in context. Would it ever make sense to report both, requiring child terms for the different continuities? When would you prefer profile over centroid or vice-versa?

psi-ms.obo
```diff
@@ -28570,6 +28573,139 @@ relationship: has_metric_category MS:4000008 ! ID based metric
...
+relationship: has_metric_category MS:4000216 ! imaging metric
+relationship: has_units UO:0000189 ! count unit
+relationship: has_value_type xsd:integer ! The allowed value-type for this CV term
+
+[Term]
+id: MS:4000226
+name: pixel grid dimensions
+def: "The dimensions, in pixels, of the smallest rectangular grid that encloses all acquired pixel coordinates of an imaging acquisition. A value pair represents the extent along x and y." [PSI:MS]
+comment: The values count pixel positions, not physical distance. A position is counted whether or not a spectrum was acquired at it.
+is_a: MS:4000004 ! n-tuple
```
mobiusklein, line 28837: This looks like it creates a bounding box, but doesn't specify the order to specify the coordinates in. This leads to ~24 different ways to write the same thing. Should a canonical order (e.g. sorted by x then y) be required? Other some of the other requested terms here seem to work in 2 or 3 dimensions. Would a bounding cube need to be a separate term?

mobiusklein (changes requested):

psi-ms.obo
patrick-bo: This uses the same wording as most n-tuples in the mzQC CV e.g. `MS:4000061`.
If it needs to be more explicit I would suggest something along the lines of "The number of values implies the quantile mode; values are the `n` inner points of the `n+1` quantiles."  Four values would correspond to the first through fourth quintile. Could add another example to the comment if desired.

psi-ms.obo
patrick-bo: Values are not directly comparable as this takes the intensity of the highest peak which differs between profile and centroid. My initial thinking was to not prescribe too much processing but fixing centroid seems like the more correct option for interpretability as well, Profile mode fragments a single peak into multiple values so a single ion taking up the majority of TIC will look severely underreported. I see no point in reporting both here, that difference is primarily a metric towards whichever software performed the centroiding.

Fix this and the corresponding matrix to centroid in the def, drop this in the note?

psi-ms.obo
patrick-bo: Yes this is a bounding box on the entire acquisition. There's some very sparse datasets (>70% of pixels contain no spectrum) and I was thinking how to include those nicely in the definition so it may read a bit odd. 

A bounding cube should be a seperate term if required, to be used us a set metric and not a run metric as 3D MSI is inherently made up of multiple acquisitions - consecutive tissue sections each with their own extent. 

Fix the notation to `(X, Y)` more explicitly in the def.

psi-ms.obo
patrick-bo: additonal TODO: directly refer to the matrix by term number as with the other n-tuples here.

mobiusklein: Thank you. If it is already pervasive, then no need to change it. I didn't find a definition of "mode" in this context looking in a few places, but I don't deal with this often.

psi-ms.obo
mobiusklein: That satisfies me, thank you.

psi-ms.obo
mobiusklein: Would it make more sense to define a bounding *n*-ary polygon in addition or is that too complex for your use-case? That would also nicely solve the order constraint because you could use path/vertex notation to draw the polygon then. This might just be the result of spending too much time breathing unfiltered air in the presence of SVGs.

patrick-bo: This is a bit too complex for this usecase which is primarily mapping the extent of the acquisition, additional benefit of the `(X,Y)` tuple is a quick sanity check of the matrices, those should have the same dimensions.
The question this metric answers is "How big of an area does the measurement cover" also allowing for easy conversion to physical distance units with consideration of pixel to pixel distance metadata.

mobiusklein: Understood, I'll stop making it complicated.

psi-ms.obo
```diff
@@ -28703,6 +28706,142 @@ relationship: has_value_type xsd:int ! The allowed value-type for this CV term
...
+relationship: has_metric_category MS:4000009 ! ID free metric
+relationship: has_metric_category MS:4000012 ! single run based metric
+relationship: has_metric_category MS:4000227 ! imaging metric
+relationship: has_units UO:0000191 ! fraction
+relationship: has_value_type xsd:float ! The allowed value-type for this CV term
+
+[Term]
+id: MS:4000235
+name: QC reference m/z
+def: "The theoretical m/z of the mass reference ion used by a mass reference based QC metric." [PSI:MS]
```
jonathan-hunter: Should this be "A theoretical m/z....." - this kind of implies an imaging acquisition can have only one reference m/z?

jonathan-hunter: If so, also check the mentions of QC reference m/z above.

psi-ms.obo
```diff
@@ -28703,6 +28706,142 @@ relationship: has_value_type xsd:int ! The allowed value-type for this CV term
...
+comment: Reported in the metadata of the runQuality or setQuality whose metrics it parameterises. REQUIRED for interpreting reference m/z error image (MS:4000232) and its quantiles (MS:4000233).
+is_a: MS:4000080 ! QC non-metric term
+
+[Term]
+id: MS:4000236
+name: duplicate pixel coordinate count
+def: "The number of distinct x/y/z pixel coordinates in a singular imaging acquisition at which more than one spectrum of the same type was recorded." [PSI:MS]
+comment: An imaging acquisition normally records at most one spectrum per pixel coordinate. A nonzero value may indicate an export or acquisition defect, such as a stage retrace or a conversion fault, and marks positions that no matrix metric can fully represent, since a matrix cell holds only one value. Multiple spectra of different types may occur in dual polarity or MS2 acquisition methods.
```
jonathan-hunter: `A non-zero value other than one`??

jonathan-hunter (approved):
Looks good to me, a couple of minor nitpicks you can consider.

psi-ms.obo
patrick-bo: > Should this be "A theoretical m/z....." - this kind of implies an imaging acquisition can have only one reference m/z?

This is somewhat of a technical limitation with the matrix format at the moment. We cannot have multiple of the same accession and cannot fit multiple values inside the matrix. If more than one reference is needed multiple runquality sections could be used but I have it on my agenda for a v1.1 revision of the specification.

mobiusklein (approved):

jonathan-hunter (changes requested):
please add `subset: mzQC` to your new terms in the MS:4n series, thanks!

patrick-bo:
> please add `subset: mzQC` to your new terms in the MS:4n series, thanks!

done, seems I missed that while skimming the new readme, my bad.

PR #553, opened 2026-09-11
Update semver regex #542

### LINKED ISSUE #542: semantic version regexp isn't properly escaped
the back slashes in the semantic version regexp should be escaped 
https://github.com/HUPO-PSI/psi-ms-CV/blob/master/psi-ms.obo#L22774
(compare to other regexp's like https://github.com/HUPO-PSI/psi-ms-CV/blob/master/psi-ms.obo#L22823)

also, there are suggested regexp's for semantic versions at https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string

Claude is suggesting:
`def: "^v?(0|[1-9]\\d*)\\.(0|[1-9]\\d*)\\.(0|[1-9]\\d*)(?:-((?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\\.(?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\\+([0-9a-zA-Z-]+(?:\\.[0-9a-zA-Z-]+)*))?$" [PSI:PI, https://semver.org/]
`
as an equivalent of the semver.org regexp that would still work in Java.
It says that consequences of using this would be:
> 1.2.3, v1.2.3, 1.2.3-alpha.1, 1.2.3+build.7, 1.2.3-alpha+b.7 **pass**; 
> 01.2.3, 1.2.3-, crosslinking-1.0.0 **fail**.

douweschulte:
Closes #542.
As discussed in the PSI meeting on 11 september.

### DIFF
--- a/psi-ms.obo
+++ b/psi-ms.obo
@@ -22772,7 +22772,7 @@ is_a: MS:1003123 ! Bruker Daltonics timsTOF series
 [Term]
 id: MS:1003384
 name: semantic version regexp
-def: "v?(\d+)\.(\d+)\.(\d+)(?:-(\S+))?" [PSI:PI]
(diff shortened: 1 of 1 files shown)


mobiusklein (approved):
Tested, looks good for valid semvers, rejects invalid ones I tried manually.

edeutsch (approved):

PR #556, opened 2026-09-14
Fix undefined subsets in psi-ms.obo

### LINKED ISSUE #543: if obo format is 1.2 then subsets need to be defined
if the obo format is version 1.2 (https://github.com/HUPO-PSI/psi-ms-CV/blob/master/psi-ms.obo#L1)
then the obo file format specification (https://owlcollab.github.io/oboformat/doc/GO.format.obo-1_2.html) says:

> subset
> [...] The value of this tag must be a subset name as defined in a subsetdef tag in the file header. If the value of this tag is not mentioned in a subsetdef tag, a parse error will be generated. [...]

I think that would mean adding to the file:
```
subsetdef: NCIT:C105763 "NCI Thesaurus subset NCIT:C105763"
subsetdef: unit_group_slim "unit group slim"
subsetdef: unit_slim "unit slim"
```
Without this some libraries won't work, specifically uk.ac.ebi.ols:ols-core v1.19 which is a dependency of the mzidentml semantic validator.

jonathan-hunter:
Fixes #543.

Remove the sole `subset: NCIT:C105763` annotation and add header definitions for the retained `unit_group_slim` and `unit_slim` subsets, as agreed in the issue discussion. All imported NCIT terms and other term content are preserved. Update the version to 4.1.262 and refresh the date and saved-by headers as required by the repository.

Validation: `check_sorted.py` (including fastobo parsing), `check_version_uprev.py`, and `git diff --check` passed. Verified that every used subset has a header definition, stanza IDs are unique, and term bodies differ only by the requested annotation removal.


edeutsch (approved):

jonathan-hunter:
TODO: Migrate changes to `psi-ms-core.obo`.

PR #557, opened 2026-09-15
new terms for GC-MS QC mix

cbielow:
A set of terms to describe a QC-Mix (we use it for GC-MS, but could be used for LC as well), mostly the retention time deviations and observed intensities.
All are single-run metrics, which makes annotation of a new QC mix independent of previous ones. For downstream comparison, multiple mzQC files can be combined.

I left a gap in the identifiers to fit in #550 before this PR (not sure if this helps).


jonathan-hunter:
Hi Chris, please migrate changes to psi-ms-core.obo - the new target for CV edits. Thanks!!

jonathan-hunter:
Also, please add `subset: mzQC` to your new terms in the MS:4n series, thanks!

cbielow:
done. Numbering is still dependent on #550

PR #567, opened 2026-09-19
fix: add missing subset detection

mobiusklein:


scripts/check_subsets.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+def get_subset_for_id(id: PrefixedIdent):
+    if id.prefix == 'PEFF':
+        return "PEFF"
+    elif id.prefix != 'MS':
+        return None
+    v = str(id.local)[0]
+    if v == '4':
+        return "QC"
```
jonathan-hunter, line 22:
```suggestion
        return "mzQC"
```

scripts/check_subsets.py
```diff
@@ -0,0 +1,46 @@
...
+
+def check_term(term) -> bool:
+    for clause in term:
+        if isinstance(clause, SubsetClause):
+            return True
+    return False
+
+
+def get_subset_for_id(id: PrefixedIdent):
+    if id.prefix == 'PEFF':
```
jonathan-hunter, line 16: Maybe also handling for NCIT, UO terms?

scripts/check_subsets.py
```diff
@@ -0,0 +1,46 @@
+import sys
+
+import fastobo
+from fastobo.term import SubsetClause
+from fastobo.id import PrefixedIdent
+
+
+def check_term(term) -> bool:
```
jonathan-hunter, line 8: Seeing as `get_subset_for_id()` has the subset value check logic already, maybe we could expand this to check that the subset/one of the subsets is of the correct value?

scripts/check_subsets.py
jonathan-hunter: Also MS:0000000 is currently unhandled but its probably not a big deal.

psi-ms-core.obo
```diff
@@ -29358,6 +29358,7 @@ is_a: MS:1000513 ! binary data array
 id: MS:1003871
 name: Lutefisk
 def: "Lutefisk is a program for the de novo interpretation of peptide CID spectra." [PSI:MS, DOI:10.1385/MB:22:3:301, DOI:10.1021/ac001196o, DOI:10.1385/1-59259-045-4:41, DOI:10.1002/(SICI)1097-0231(19970615)11:9<1067::AID-RCM953>3.0.CO;2-L, https://sourceforge.net/projects/lutefiskxp/, https://www.hairyfatguy.com/lutefisk/]
+subset: core
```
jonathan-hunter, line 29361: 😆

jonathan-hunter (changes requested):

